PR #6, opened 2022-07-01
[firmware] add minimum platform requirement

ahinrichs:
Firmware needs espressif32 platform version 4


tbnobody:
Thank you for this PR

PR #31, opened 2022-07-15
[webapp] close navbar when item is clicked

ahinrichs:
I haven't found another way[1] but to manually remove the `show` class with onClick. 

-- 
[1] except switching to bootstrap-vue...


tbnobody:
Imho bootstrap-vue is no option as it doesn't support Bootstrap 5 and Vue 3.

But wouldn't it be enough to add `data-bs-toggle="collapse" data-bs-target="#navbarNavAltMarkup"` to the links?

ahinrichs:
Tried that. The NavBar got collapsed but the link wasn't followed. I guess the generated JS does not propagate the click to the Router / A-Tag.

ahinrichs:
Wait a second. When I tried to reference my source I found one last different approach.... Checking

Issue #83, opened 2022-08-16
OpenDTU messing up configuration during runtime


HacksBugsAndRockAndRoll:
During normal runtime on version https://github.com/tbnobody/OpenDTU/commits/0cc6ce3 the inverter configuration seems to break.

![image](https://user-images.githubusercontent.com/61069044/184896345-cdd8af30-4848-4448-86cd-d2928babfe3c.png)

curl gives this info /serial changed, but it is the correct one)
>curl http://192.168.178.96/api/inverter/list
>{"inverter":[{"id":0,"name":"SoxSolar","serial":"114181XXXXX","type":"Unknown","max_power":[0,0,0,0]}]} 

mqqt publishes like this
![image](https://user-images.githubusercontent.com/61069044/184898881-a149434c-7dfd-4eb2-ba99-59e256692de2.png)

OpenDTU no longer pulls data from the inverter in this state.

tbnobody:
Could you please doublecheck the heap usage on the system overview if this error occours again?

HacksBugsAndRockAndRoll:
![image](https://user-images.githubusercontent.com/61069044/184981532-fb0cfe83-f022-4f24-87f3-37568a63d34c.png)

![image](https://user-images.githubusercontent.com/61069044/184981960-be0a843e-4ba0-44cf-9e4a-5cbb6f946eb1.png)

also the dtu is currently still in this state. Is there anything else I should look into?

tbnobody:
The Heap usage look a little bit high, but not too much. I am currently seeing something between 108 and 110kb.

You mentioned in the other issue that this only occours for one of your two ESP's. Have you tried to reflash this one again (After a complete flash erase)? 

Do you see anything special in the serial consoole before this issue happens?

HacksBugsAndRockAndRoll:
I opened the ticket because I thought it was the ESP hardware but now the second ESP also shows this behavior. 
Both ESPs were flash erased ~3 days ago. 
The USB is only connected to power for this ESP at the moment since I cannot get the inverters signal at my PC.
I monitored the heap usage by polling ~30s see here (captured some spikes but always seems to recover fine)
![image](https://user-images.githubusercontent.com/61069044/184990799-62a93687-470a-4a6b-a6a4-2641c69d4b5a.png)

tbnobody:
Today, before I flashed a new version, the uptime was > 2 days without any issues. So you maybe have some kind of different config. Are you using MQTT with TLS?

[...]

PR #93, opened 2022-08-19
fix static code analysis warnings

HacksBugsAndRockAndRoll:
still looking for the culprit behind https://github.com/tbnobody/OpenDTU/issues/83 

some findings from cppcheck fixed

- lots of initialization stuff
- overrides
- const for buffer parameters

actual logic changes
- in lib/Hoymiles/src/parser/AlarmLogParser.cpp
- all the places printing the inverter serial as hex string


stefan123t:
@HacksBugsAndRockAndRoll how did you use cppcheck, may we add this easily to the github actions ?
Some of the findings may be useful initializations, easier syntax, etc.

tbnobody:
Will close this PR as a lot of code has already changed.

PR #94, opened 2022-08-19
src/Configuration.cpp: fix homeassistant discovery config

madmartin:
On a fresh installed openDTU esp32, the mqtt discovery prefix
topic does not match the default value of homeassistant, due to a wrong
assignment in src/Configuration.cpp.
This patch fixes it.

Signed-off-by: Martin Dummer <martin.dummer@gmx.net>

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -54,7 +54,7 @@ void ConfigurationClass::init()
 
     config.Mqtt_Hass_Enabled = MQTT_HASS_ENABLED;
     config.Mqtt_Hass_Retain = MQTT_HASS_RETAIN;
-    strlcpy(config.Mqtt_Hass_Topic, MQTT_TOPIC, sizeof(config.Mqtt_Hass_Topic));
+    strlcpy(config.Mqtt_Hass_Topic, MQTT_HASS_TOPIC, sizeof(config.Mqtt_Hass_Topic));
     config.Mqtt_Hass_IndividualPanels = MQTT_HASS_INDIVIDUALPANELS;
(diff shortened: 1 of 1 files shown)


HacksBugsAndRockAndRoll (approved):

maxencep02:
Hello,
I have done the little modification and recompile the project. HomeAssistant still doesn’t autodetect my DTU. Have you done any other modification? 
I must said that it is a fresh installation from scratch because I received the component yesterday.

HacksBugsAndRockAndRoll:
> 

several things need to be ensured:

HA:
 - mqtt auto discovery needs to be enabled in home assistant ( https://www.home-assistant.io/docs/mqtt/discovery/#configuration )
 - also make sure note of your discovery_prefix

DTU:
image

make sure this matches your discovery_prefix
image

this worked for me before this fix, I just had to manually adjust the prefix which should now be preconfigured to the HA default

maxencep02:
Thank you to point out the issue. The HASS Prefix Topic had remain on "solar/". After changing it it is autodiscovered.

madmartin:
> Thank you to point out the issue. The HASS Prefix Topic had remain on "solar/". After changing it it is autodiscovered.

Please take note that the config is stored separately from the code.
This pull request fixes the behavior for the  **initial config** of new installations (like after a flash erase on your esp32 and flash the application again). When you updated the code, build and flashed it **without** erasing your esp32 before, the previous config still exists and so there is still "solar/" stored and nothing has changed in your config. So its normal that you must change it manually in this case.

PR #102, opened 2022-08-22
src/NetworkSettings.cpp: create unique hostname

madmartin:
the initial config contains a default hostname "OpenDTU-%06X" where
the characters "%06X" are supposed to be replaced with the last 3 bytes
of the macaddress in hex
This patch implements the creation of the unique hostname once, it is
stored as part of the configuration

Signed-off-by: Martin Dummer <martin.dummer@gmx.net>


HacksBugsAndRockAndRoll (commented):
Instead of string replace you could probably use the default hostname as a format string in a sprintf.. Looks like this was the original intent

madmartin:
@tbnobody reworked the patch

PR #103, opened 2022-08-23
Make hoymiles library usable for other architectures

greenflash1986:
Hi @tbnobody first of all thanks for the great work.

After I were unable to log my supply from my inverter with the Voltcraft plug, I stumpled accross the thread on Mikrocontroller. The work done there is very appreciated and although I'm unable to remember all the posts I where able to find the two repositories.

Now I want to use my spare Atmega328PB for the logging and I found your repository with the library approach the most promising. Although I don't know if the Atmega has enough speed or RAM (currently the Atmega-branch with only a simple main builds with >4kb and the Atmega only has 2kb) I already did some things to make the hoymiles library compile for the `atmelavr`-toolchain.

This PR is a request for comments: it's an idea to make things working for multiple platforms and maybe you want to change the library code in such a manner. I'm a software developer but my embedded skills and C++ knowledge is very limited. Therefore I would not assume that my code is correct in all details.
Please have a look and feel free to comment.

Best regards
GreenFlash1986


greenflash1986:
Updated base to current master.

Best regards

Sprinterfreak (commented):
Multiple radio requirement is on the horizon.
Currently reverse engineering of HMS series inverters is ongoing, which require different radio hardware.

stefan123t:
@greenflash1986 thanks for your efforts. Given that OpenDTU is targeted for the ESP32 architecture/platform only at the moment, not even ESP8266, I do not see this PR going anywhere soon. Though it may serve as an exercise for the future.

PR #114, opened 2022-08-31
Option for expire_after

battifreund:
Added UI-Option for HA Auto Discovery attribute "expire_after". For some use cases it is more convinient not to set an expiration time.


tbnobody:
Hint: You should never ever place new config settings in the middle of the struct. Otherwise all following settings get assigned wrong.

battifreund:
> Hint: You should never ever place new config settings in the middle of the struct. Otherwise all following settings get assigned wrong.


Yes, of course. My mistake! I just found out too. I think my need for order got the better of me.

Sorry!

I also noticed one thing: does the JSON buffer (MQTT_JSON_DOC_SIZE) need to be increased? It is currently set to 3072 bytes.

madmartin:
Hello @battifreund 
good work!
One technical sidenote to your pull request:
If you fix your own pull request like this, before it is merged, it is very much appreciated that you do a rebase and squash your two commits into one.
And then do a "git push -f" to replace the previous commit with the new one.
Let me know if you need the list of commands to execute for this.

battifreund:
Hi @madmartin 

Thank you very much for your support. I think I've worked it out. :) 
I've done a rebase (git rebase -i origin/master) , picked my first commit and squashed the second into it. Then I've done a forced push (git push -f origin Feature-expire_after).

I hope, I've done everything right now. I'm learning :)

Better now?

include/Configuration.h
```diff
@@ -4,7 +4,7 @@
 #include <Arduino.h>
 
 #define CONFIG_FILENAME "/config.bin"
-#define CONFIG_VERSION 0x00011400 // 0.1.20 // make sure to clean all after change
+#define CONFIG_VERSION 0x00011500 // 0.1.20 // make sure to clean all after change
```
tbnobody, line 7: You should also update the version number (0.1.21) in the comment

src/MqttHassPublishing.cpp
```diff
@@ -116,7 +116,9 @@ void MqttHassPublishingClass::publishField(std::shared_ptr<InverterAbstract> inv
         root[F("unit_of_meas")] = inv->Statistics()->getChannelFieldUnit(channel, fieldType.fieldId);
         root[F("uniq_id")] = String(serial) + "_ch" + String(channel) + "_" + fieldName;
         root[F("dev")] = deviceObj;
-        root[F("exp_aft")] = Hoymiles.getNumInverters() * Configuration.get().Mqtt_PublishInterval * 2;
+        if(Configuration.get().Mqtt_Hass_Expire) {
```
tbnobody, line 119: Please add a space between the `if` and the bracket

PR #141, opened 2022-09-13
Fixed compile time error related to pervious changes (OPENDTU_ETHERNET)

kreutter:


tbnobody:
thank you for the fix

PR #161, opened 2022-09-24
Introduce hostname and dataagecritical via MQTT

MFV:
Publish the hostname beside the IP of the device via MQTT. 
In addition publish per inverter the Data Age Critical flag as used for the Web UI. This allows you to visualize the sleep phase of the inverter at night and detect outages.


tbnobody:
I've implemented this a little bit different. But should be included in 747335f23fb35a3f59fa88ff5fdc08320a05c6c8 and 4fe21111737f3617eda89e7fbbf62e7179b9d74d

PR #200, opened 2022-10-07
com angepasst

geraldriess:


tbnobody:
Sorry, it doesn't make sense to apply a pull request that changes the serial port. The port is maybe different for every user.

PR #210, opened 2022-10-11
Adding another available case

cepresso:
Thanks for the great work on OpenDTU!
I have created another 3d printable case I would like to share with everyone interested.


tbnobody:
@cepresso can you confirm that the pinout as mentioned in #308 is correct? Does it make sense to add this documentation to your printables page or should I add this to the readme as shown in the PR #308?

cepresso:
Yes, that's the pins I used on my board. I will add it to the printables page, I think it makes more sense there together with the pictures.
In the end one can choose any pins. Unfortunately, that short ESP32 board doesn't expose pin 16.

PR #304, opened 2022-11-04
BREAKING CHANGE: add dtu serial to mqtt status topics

madmartin:
### +++WIP+++

When two OpenDTU devices are connected to one mqtt broker, it is not possible
to distinguish the status topics between these two devices.

The status topics are composed like

`<base topic>/dtu/<status item>`

example:

```
solar/dtu/rssi -61
solar/dtu/uptime 395
solar/dtu/ip 192.168.166.189
solar/dtu/hostname OpenDTU-8E1C80
solar/dtu/rssi -59
solar/dtu/uptime 7472
solar/dtu/ip 192.168.166.192
solar/dtu/hostname OpenDTU-83B278
```

proposed solution:

insert dtu serial into the status topics like

`<base topic>/dtu/<dtu serial>/<status item>`

```
solar/dtu/199980113408/uptime 2838
solar/dtu/199980113408/ip 192.168.166.189
solar/dtu/199980113408/hostname OpenDTU-8E1C80
solar/dtu/199980113408/rssi -56
```

BUT:

what to do with the configurable  LWT topic `solar/dtu/status` ?

I did not touch this yet. I see two options:
- add a placeholder variable for the serial number to the configurable default, like `solar` / `dtu/#SERIAL#/status` 
- remove the LWT topic from configuration (why is this configurable anyhow?)

What do you think?

### DIFF
--- a/docs/MQTT_Topics.md
+++ b/docs/MQTT_Topics.md
@@ -4,13 +4,15 @@ The base topic, as configured in the web GUI is prepended to all follwing topics
 
 ## General topics
 
+serial will be replaced with the serial number of the OpenDTU device.
+
 | Topic                                   | R / W | Description                                          | Value / Unit               |
 | --------------------------------------- | ----- | ---------------------------------------------------- | -------------------------- |
-| dtu/ip                                  | R     | IP address of OpenDTU                                | IP address                 |
-| dtu/hostname                            | R     | Current hostname of the dtu (as set in web GUI)      |                            |
-| dtu/rssi                                | R     | WiFi network quality                                 | db value                   |
+| dtu/[serial]/ip                         | R     | IP address of OpenDTU                                | IP address                 |
+| dtu/[serial]/hostname                   | R     | Current hostname of the dtu (as set in web GUI)      |                            |
+| dtu/[serial]/rssi                       | R     | WiFi network quality                                 | db value                   |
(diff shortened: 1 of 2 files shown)


tbnobody:
It was the idea to change the base topic if you really want to use multiple devices. Downside is, that you will have your inverters on different sub topics. Don't know your exact use case.

madmartin:
So what you propose is to configure the devices like

first one:

```
solar-one/dtu/hostname OpenDTU-8E1C80
```

and the other one;

```
solar-two/dtu/hostname OpenDTU-83B278
```

?

If yes, then my opinion is: Not nice.

madmartin:
Another point: this still does not explain why the LWT (sub)topic sould be configurable. I see no benefits with that.

tbnobody:
Let's just say I required that in the past. We should maybe move the discussion regarding mqtt topics into the discussion settings. If I would change this and make It incompatible with the current installations I would do it only once. Saw on Twitter and from other people that some other things are not optimal as well.

tbnobody:
Please see https://github.com/tbnobody/OpenDTU/discussions/317

ThomasCr:
for me - I prefer to change the main topic. So pls - when you want to change something, make it possible to configure it - to make everybody happy. A good example is Tasmota - there it is possible to configure the full topic!

https://tasmota.github.io/docs/MQTT/#mqtt-topic-definition

madmartin:
close PR unmerged

PR #427, opened 2022-12-09
Automatically create "Releases"

diecknet:
Hello,
as suggested in #403: I propose to use the GitHub "Releases" feature to provide the built .bin files in a more visible way.
In summary this PR does the following:

* release-tag is generated as a step (format: `build-<short-commitid>` fetched by running `git rev-parse --short`)
* automatically tag every commit to the repo by running `git tag` and `git push`

* Removes the previously enforced manual usage of tag for release job

I tested it in a [fork of the repo](https://github.com/diecknet/OpenDTU), works well so far. Anyway if you have suggestions for improvement before merging, let me know :)

Screenshot of forked repo with releases


diecknet:
Let me know, if you're interested in merging this PR. If so, I will happily update the PR.

Dom1n1c:
Great this has some potential (auto updates and way more)

CONSULitAS:
Great idea, but I would add an ISO timestamp to the tag for readability and for the less experienced users.

.github/workflows/build.yml
```diff
@@ -101,17 +101,18 @@ jobs:
...
-        with:
-          name: opendtu-${{ matrix.environment }}
-          path: |
-            .pio/build/${{ matrix.environment }}/opendtu-${{ matrix.environment }}.bin
-            .pio/build/${{ matrix.environment }}/partitions.bin
-            .pio/build/${{ matrix.environment }}/bootloader_dio_40m.bin
-            .pio/build/${{ matrix.environment }}/boot_app0.bin
+      - name: Generate Tag name for release
+        run: |
+          echo "tag=build-$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
```
CONSULitAS, line 106: Something like 
```
echo "tag=build-$(git rev-parse --short HEAD)-$(date -I)" >> $GITHUB_OUTPUT
```

diecknet: > Great idea, but I would add an ISO timestamp to the tag for readability and for the less experienced users.

I like that! I just used the plain Commit-ID because currently the `Info -> System` page shows it as the "Firmware version".

Screenshot of Info -> System

For even more improved readability, I would swap these values. So date first, then commit id. Example:

> build-2023-02-19-da86714

diecknet:
I updated the PR to resolve the merge conflicts. 
I also adjusted the Release Tag to include the build date.

ValleBL:
Would love to see this added. Just a little bit more convenient

tbnobody:
I am going to close this PR. In the end it's better to generate the tags manually to have a chance to generate nightly builds or test builds if required..

PR #526, opened 2023-01-20
Added alternative PIN Layout of ESP32 board

theozordako:
Added alternative PIN-Layout of DEVKI TV1 ESP32 Developer board


theozordako:
Added alternative PIN Layout of ESP32 board

tbnobody:
If this is still a relevant PR, please add a PR in the OpenDTU-docs repository.

PR #594, opened 2023-02-16
Fix typos in German localisation and change some words to more common…

mozzbozz:
… ones

In German, you bascially never can have two nouns after each other, just separated by a space. Either you write them together (Dampfschifffahrtsgesellschaftskapitän) or in case one of them is a foreign word, you *can* but don't have to separate them by a hyphen (Firmware-Aktualisierung).

Additionally, I've changed some more details like WiFi -> WLAN, because that is a lot more common in German. I've also removed the "-Einstellungen" from the settings dropdown, because having it on every single item just makes it less readable (especially considerung the dropdown itself already is called "Einstellungen", so it's clear, all those items are for settings) (now it does work out the same way as the Info dropdown)


Dom1n1c:
But WiFi is technically more accurate, there is a WiFi Alliance, bit nothing about WLAN 📶

CONSULitAS:
Maybe, but in Germany nobody uses the term Wi-Fi in common language. WLAN is the common term. 

Just look at this iOS screenshot for example:
IMG_0984

dAjaY85:
https://www.heise.de/tipps-tricks/WiFi-oder-WLAN-was-ist-der-Unterschied-4571128.html

CONSULitAS:
> There is a difference

Well known to us nerds. But if you scroll down in this article you find:

IMG_0985

And which term does even Heise use in their headlines? WLAN without exception. 🤣🤣🤣

mozzbozz:
@Dom1n1c well... per the article someone else linked, every WiFi is a WLAN anyways. WLAN is just a more general word. So while less specific, it's still technically correct - therefore, I think the more commonly used / more intuitive word in a language should be used.

PR #603, opened 2023-02-19
fix typo: publish_invterval

morph027:


tbnobody:
Unfortunatly it's not possible to just change this value. It would break all users installations and reset the value to the default. Have to implement a migration.

morph027:
Ah, of course. I see. Any clue how to do this? I'm more the python guy 😀

tbnobody:
Yes... will commit it in the next days...

morph027:
Are you planning to include the fix? If so, we can just close this PR then.

tbnobody:
Implemented in dc66a82695d37e6c15c38a91a26a4cf9cb51e9fd

PR #624, opened 2023-02-24
Fix prometheus counters

### LINKED ISSUE #616: Prometheus metric `opendtu_YieldTotal` (and `opendtu_YieldDay`) should be counter, not gauge (and follow counter semantics strictly)
### What happened?

Among the prometheus metrics exposed by OpenDTU, there are also yield metrics (total and per day). 

```
# HELP opendtu_uptime Uptime in seconds
# TYPE opendtu_uptime counter
opendtu_uptime 8462
...
# HELP opendtu_YieldDay in Wh
# TYPE opendtu_YieldDay gauge
opendtu_YieldDay{serial="123",unit="0",name="BKW",type="AC",channel="0"} 70.000000
# HELP opendtu_YieldTotal in kWh
# TYPE opendtu_YieldTotal gauge
opendtu_YieldTotal{serial="123",unit="0",name="BKW",type="AC",channel="0"} 4.666000
```

While it is arguable, whether the daily yield is more of a `counter` or a `gauge` (or if it is a useful metric for prometheus at all, since it has per-day-semantics), the total Yield should be a `counter`, since it is meant as a eternal sum, monotonously rising (maybe resetting to 0 after certain events, e.g. inverter firmware updates).

### To Reproduce Bug

Open http://hostname.of.opendtu/api/prometheus/metrics 
Search for `# TYPE opendtu_YieldTotal`
It is declared as type `gauge`

### Expected Behavior

`opendtu_YieldTotal` should be declared as a `counter`, like `opendtu_uptime`.
Probably the same is valid for `opendtu_YieldDay`.

### Install Method

Self-Compiled

97bc964

### Relevant log/trace output

_No response_

### Anything else?

https://prometheus.io/docs/concepts/metric_types/

easimon:
This PR does two things

* declare counter-type prometheus metrics as such (yield and eventlog entries are monotonously rising, occasionally resetting to 0)
* after a reboot of the DTU, do not report inverter metrics before first successful read from the inverter

I tested it by rebooting the DTU during daytime, while the inverter was on, and also rebooted the DTU over night.
In both cases, the first few metrics scrapes did not contain any inverter statistics before the very first contact with the inverter. Also, the counter metrics are now declared as such.

This fixes #616.

Rationale: 

## Metric types

Prometheus differentiates between counters and gauges (and others not used here). Quote from https://prometheus.io/docs/concepts/metric_types/ :

> The Prometheus client libraries offer four core metric types. These are currently only differentiated in the client libraries (to enable APIs tailored to the usage of the specific types) and in the wire protocol. The Prometheus server does not yet make use of the type information and flattens all data into untyped time series. This may change in the future.

So, for correctness, the metrics that are counters, should be declared as such.

## Not reporting before successful update

Prior to this change, OpenDTU reported `0` after a reboot, for all metrics that had not yet been updated. This was the case until the first contact with the inverter. If that reboot happens over night, the first contact occurs only hours later, at sunrise, since the inverter shuts down over night. Reporting `0` in counters when it really means "no information" breaks some prometheus functions like [rate()](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate), [increase()](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase), since these consider this a counter reset. If you don't know a value, they should not be reported at all, so the built-in reset-behaviour of these functions works correctly.
Prometheus can handle intervals with no values -- but not intervals with incorrect values. 
Example: 

The daily yield counter is monotonously increasing, but resets every night. So a time series for it might look like

... 200, 201, 221, 0, 20, 25, 40 ...

between 221 and 0, the counter was reset. Using increase(), one can calculate the total yield over the interval, which would be (221-200) + (40-0) = 61. You can also select arbitrary sub-intervals of the time series, with multiple resets.

Note that the Hoymiles inverter does no resets on restarts, e.g. the total yield counter increases eternally. Especially for this counter, reporting 0s inbetween would lead to drastic miscalculations.

... 2000, 2001, 2020, (dtu reboot), 0, 0, 0, (first contact), 2030, 2040, 2050, (dtu reboot), 0, (first contact), 2060, 2070 ...

Using increase() over the total interval, prometheus would return (2020-2000) + (2050-0) + (2070-0) = 4140, when it really should be 2070-2000 = 70.

With this change, the time series looks like this

... 2000, 2001, 2020, (dtu reboot), (no values), (first contact), 2030, 2040, 2050, (dtu reboot), (no values), (first contact), 2060, 2070 ...

and then increase() would calculate 2070-2000 = 70, since there was no reset inbetween.

### DIFF
--- a/lib/Hoymiles/src/parser/StatisticsParser.cpp
+++ b/lib/Hoymiles/src/parser/StatisticsParser.cpp
@@ -130,6 +130,12 @@ const char* StatisticsParser::getChannelFieldName(ChannelType_t type, ChannelNum
     return fields[pos->fieldId];
 }
 
+const char* StatisticsParser::getChannelFieldMetricType(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId)
+{
+    const byteAssign_t* pos = getAssignmentByChannelField(type, channel, fieldId);
+    return field_metric_types[pos->fieldId];
+}
+
 uint8_t StatisticsParser::getChannelFieldDigits(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId)
 {
     const byteAssign_t* pos = getAssignmentByChannelField(type, channel, fieldId);
--- a/lib/Hoymiles/src/parser/StatisticsParser.h
+++ b/lib/Hoymiles/src/parser/StatisticsParser.h
@@ -43,6 +43,9 @@ enum FieldId_t {
 const char* const fields[] = { "Voltage", "Current", "Power", "YieldDay", "YieldTotal",
     "Voltage", "Current", "Power", "Frequency", "Temperature", "PowerFactor", "Efficiency", "Irradiation", "ReactivePower", "EventLogCount" };
 
+const char* const field_metric_types[] = { "gauge", "gauge", "gauge", "counter", "counter",
+    "gauge", "gauge", "gauge", "gauge", "gauge", "gauge", "gauge", "gauge", "gauge", "counter" };
+
 // indices to calculation functions, defined in hmInverter.h
 enum {
     CALC_YT_CH0 = 0,
@@ -103,6 +106,7 @@ public:
     bool hasChannelFieldValue(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId);
     const char* getChannelFieldUnit(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId);
     const char* getChannelFieldName(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId);
+    const char* getChannelFieldMetricType(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId);
     uint8_t getChannelFieldDigits(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId);
 
     float getChannelFieldOffset(ChannelType_t type, ChannelNum_t channel, FieldId_t fieldId);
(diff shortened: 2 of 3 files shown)


tbnobody:
Please place the field_metric_types variable in the prometheus files. These values have nothing todo with hoymiles communication. Therefor please not into the Hoymiles lib. 
Something similar is already implemented in the HomeAssistant integration.

easimon:
I see your point. But I am not sure about how to do this in a maintainable fashion (maybe also missing C++ skills).

1. There is an implicit relationship between `fields` and `field_metric_types`, they must be of same length and order, because each `field_metric_types[i]` is connected to `fields[i]`.
2. Also I did not completely follow the logic of resolving the correct array index based on `getAssignmentByChannelField(type, channel, fieldId)`, which then also should move/be replicated in the Prometheus class, I guess.

When doing so, this might be harder to refactor later on -- or simply break unnoticed if the fields are changed (not sure if that can happen, since the Hoymiles protocol will hopefully not change drastically). But if that happens, one needs to change both field_metric_types and fields symmetrically, and this is easier to slip through when the variables are not at least close to each other.

Do you have a hint on how to refactor this in a good way?

tbnobody:
You could create a mapping between field type and prometheus type like here:
https://github.com/tbnobody/OpenDTU/blob/97bc964b6c761401cbd72120e1db564a78cdb91b/include/MqttHandleHass.h#L33-L49

tbnobody:
I've just written  it using a map... will commit it soon

tbnobody:
Should be fixed  with  c67b3e3cbdf944786b53001f98e82e217dc8cc89

PR #637, opened 2023-02-27
Add alternative Hardware Settings PIN names

tabacha:
My NRF24+ is labeled like this. 
image
So I would suggest to add this altenative names to the Web GUI.

### DIFF
--- a/webapp/src/locales/fr.json
+++ b/webapp/src/locales/fr.json
@@ -512,6 +512,7 @@
         "Category": "Catégorie",
         "Name": "Nom",
         "ValueSelected": "Sélectionné",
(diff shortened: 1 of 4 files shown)


webapp/src/locales/fr.json
```diff
@@ -1,517 +1,518 @@
-{
```
tabacha, line 1: This file was encoded with CR+LF encoding. I changed it to unix style as in en.json and de.json

tbnobody:
Unfortunatly this cannot be applied because the pin table is no generated automatically based on the json structure

tabacha:
Isnt it generarted by PinInfo.vue ?

PR #728, opened 2023-03-21
hoymiles: keep backbuffer for StatisticsParser, swap only on success [fixes #727]

### LINKED ISSUE #727: Prometheus - Werte sind manchmal 0
### What happened?

Falls openDTU ein unvollständiges Paket erhält kann es passieren das alle Prometheus Werte 0 since.

Das führt zu unschönen Zacken und Problemen in den Daten, opendtu_last_update wird fleißig hochgezählt obwohl die Daten nicht aktualisiert wurden.

Screenshot 2023-03-21 at 13 54 12
Screenshot 2023-03-21 at 13 54 28

### To Reproduce Bug

System laufen lassen mit und die /api/prometheus/metrics page öfters refreshen, irgendwann sind alle Werte 0, falls RX Fragments fehlen. (>20m Distanz zum WR)

### Expected Behavior

Prometheus Werte sollten entweder nicht aktualisiert werden, oder gar nicht ausgegeben werden.

### Install Method

Self-Compiled

bf05dc8

### Relevant log/trace output

_No response_

### Anything else?

_No response_

Thoro:
Updates the statistics parser so that only valid statistics packets are "published" (swapped).

Fixes #727

### DIFF
--- a/lib/Hoymiles/src/parser/StatisticsParser.cpp
+++ b/lib/Hoymiles/src/parser/StatisticsParser.cpp
@@ -35,8 +35,19 @@ void StatisticsParser::setByteAssignment(const std::list<byteAssign_t>* byteAssi
 
 void StatisticsParser::clearBuffer()
 {
-    memset(_payloadStatistic, 0, STATISTIC_PACKET_SIZE);
+    memset(_payloadStatistic_off, 0, STATISTIC_PACKET_SIZE);
     _statisticLength = 0;
+}
+
+// If the statistics packet has the full size, we swap and confirm that the data is good (return true)
+bool StatisticsParser::swap()
+{
+    if (_statisticLength == STATISTIC_PACKET_SIZE) {
+        memcpy(_payloadStatistic, _payloadStatistic_off, STATISTIC_PACKET_SIZE);
+        return true;
+    }
+
+    return false;
 }
 
 void StatisticsParser::appendFragment(uint8_t offset, uint8_t* payload, uint8_t len)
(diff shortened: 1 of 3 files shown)


lib/Hoymiles/src/parser/StatisticsParser.cpp
```diff
@@ -35,17 +35,28 @@ void StatisticsParser::setByteAssignment(const std::list<byteAssign_t>* byteAssi
...
 {
-    memset(_payloadStatistic, 0, STATISTIC_PACKET_SIZE);
+    memset(_payloadStatistic_off, 0, STATISTIC_PACKET_SIZE);
     _statisticLength = 0;
 }
 
+// If the statistics packet has the full size, we swap and confirm that the data is good (return true)
+bool StatisticsParser::swap()
+{
+    if (_statisticLength == STATISTIC_PACKET_SIZE) {
```
tbnobody, line 45: You cannot just compare to STATISTIC_PACKET_SIZE  as each inverter sends other amount of bytes. If the program flow reaches the handleResponse method, for sure, all packages are received. At least if the `handleResponse` method returns true. But it might be that the webserver callback occurs between the copying the fragments using the `appendFragment` method.

Thoro: I checked regarding the webserver callback, first I implemented the backbuffer but the issue was still there, so it's not a timing issue.

Reg. Packet size, I guess we'll need to use the size from the inverter then.  Is the packet size for each inverter available somewhere?

PR #738, opened 2023-03-23
Hint for group add in arch in readme

Kraego:


.gitignore
```diff
@@ -7,3 +7,4 @@
 platformio-device-monitor*.log
 platformio_override.ini
 .DS_Store
+webapp_dist
```
tbnobody, line 10: Will not add this as it's makes things hard to distribute pre-compiled binaries.

README.md
```diff
@@ -199,7 +199,7 @@ esptool.py --port /dev/ttyUSB0 --chip esp32 --before default_reset --after hard_
 Users report that [ESP_Flasher](https://github.com/Jason2866/ESP_Flasher/releases/) is suitable for flashing OpenDTU on Windows.
 
 #### Flash with [ESP_Flasher](https://espressif.github.io/esptool-js/) - web version
-It is also possible to flash it via the web tools which might be more convenient and is platformindependent.
+It is also possible to flash it via the web tools which might be more convenient and is platform agnostic.
```
tbnobody, line 202: What do you mean with agnostic in this context? platform independent (ok, have to add a space) is a very common phrase.

README.md
```diff
@@ -144,7 +144,7 @@ You can also change  the pins by creating a custom [device profile](docs/DeviceP
     * upload_port
     * monitor_port
 * Select the arrow button in the blue bottom status bar (PlatformIO: Upload) to compile and upload the firmware. During the compilation, all required libraries are downloaded automatically.
-* Under Linux, if the upload fails with error messages "Could not open /dev/ttyUSB0, the port doesn't exist", you can check via ```ls -la /dev/tty*``` to which group your port belongs to, and then add your user this group via ```sudo adduser <yourusername> dialout```
+* Under Linux, if the upload fails with error messages "Could not open /dev/ttyUSB0, the port doesn't exist", you can check via ```ls -la /dev/tty*``` to which group your port belongs to, and then add your user this group via ```sudo adduser <yourusername> dialout``` (if you are using ```arch-linux``` use: ```sudo gpasswd -a <yourusername> uucp```, this method also requires a reboot).
```
tbnobody, line 147: Reboot is not required... just a logout / login of the affected user

Kraego: Changed

README.md
Kraego: It basically means the same https://english.stackexchange.com/a/12903 ... changed it back anyway, was just opinion based

.gitignore
Kraego: removed ignore

PR #756, opened 2023-03-27
Allow MQTT publishing every second

jstammi:
Consistent with DTU settings for polling data


jstammi:
Making this possible does not mean you have to use it like this in every constellation.

With a single Inverter it perfectly makes sense if you do not want to miss data. At the moment I poll and publish with both set to 2s. With publishing only at current minimum of 5s, only every 2nd-3rd sample polled could be used on listening for mqtt values.

And with multiple inverters (but even for my single-one-setup) ... I consider sending just the data of the one having polled from latest would be the most obvious expected default behaviour ... ?

PR #786, opened 2023-04-09
NavGuard

florian-senn:
Unauthorized users only can visit MainPage and Login


tbnobody:
But I want non-authorized users to visit the info views. 
If you don't want this, just deactivate this setting:
image


On the other hand, the backend is still readable with your PR. Just the frontend prevents the route access.

florian-senn:
Unter der Route "/" ist ja die Live-Ansicht, diese ist dadurch noch zugänglich. Aber ja, es schützt nur das Frontend.

3DJupp:
Hey, I mean, thats always something that should be taken care of. I recommend to use Cloudflare Access or any other kind of reverse proxy for remote access, so the ESP will not host anything in the web. For accessing the ESP in a dedicated VLAN i could allow or block certain protocols and ports.

florian-senn:
I've to take a closer look, as I haven't followed recent development further :-)

PR #801, opened 2023-04-12
Feature: MQTT add TLS authentication

madmartin:
New Feature:
User asked for TLS client certificate based login from DTU to MQTT
server. This PR implements storage and use of x509 client certificate
and private key.

A good guide for testing with mosquitto is http://www.steves-internet-guide.com/creating-and-using-client-certificates-with-mqtt-and-mosquitto/


madmartin:
See #780

PR #802, opened 2023-04-12
Leave AP opened if we allow read-only mode

mathieucarbou:


tbnobody:
I don't see what the read-only setting has todo with an open wifi. 
An always open wifi can easily acheaved using this setting:
image

mathieucarbou:
If someone setups his DTU with read-only mode activated, when the network is not available and OpenDTU reboots in AP mode, I don't want to have to put a password. I have setup read-only mode to allow someone to have access (in read-only) to the DTU. He should also have access in read-only mode to the DTU when connected to its WiFi, and without any password since read-only mode has a concept already of "public" and not private.

Sorry I didn't fully explained my idea behind that, and so you missed the point I think ;-)

PR #805, opened 2023-04-13
fix typo

mpw96:


webapp/src/locales/en.json
```diff
@@ -418,7 +418,7 @@
         "Status": "Status",
         "Send": "Send",
         "Receive": "Receive",
-        "StatusHint": "<b>Hint:</b> The inverter is power by it's DC input. If there is no sun, the inverter is off. Requests can still be sent.",
+        "StatusHint": "<b>Hint:</b> The inverter is powered by it's DC input. If there is no sun, the inverter is off. Requests can still be sent.",
```
jhenninger, line 421: And `it's` should be `its`  ;)

mpw96: Thank you, done.

PR #808, opened 2023-04-16
Feature: Admin AccessPoint Timeout now configurable

madmartin:
Make the administrative accesspoint timeout configurable.
The default value is 3 minutes, values from 0-99999 are possible, where 0 means infinite (no timeout).


tbnobody:
Merged in 8bfa7e530cfdd1bab1455c523ebb0a66a93c8b20

PR #856, opened 2023-04-28
Fix CMT SPI driver and SPI HW allocation and provide openDTUFusion build option

markusdd:
This PR adresses an issue in the CMT SPI driver.

Although Half-Duplex mode is used, the register read funtion utilized both  tx and rx buffers.
The TX buffer though shoudl only be used for data, not for providing the cmd and addr bits.
In ESP32 Framework this is still permitted and works by accident, although not intended this way.
In all later chips S2/S3/C3 etc. this leads to runtime assert and crashes openDTU. (this also means if anyone with an S2/S3/C3 tried to configure the CMT it would crash their installation)

This is fixed now by changing the setup of the spi reg device and using the cmd and addr bits as intended.
Also, the Octa-SPI flag was removed in the FIFO write function. This was non-sensical (and would crash in S2/S3/C3) as the CMT is a single data line SPI device, not an Octa-SPI with 8 parallel data lines. 

Also, this PR fixes the SPI device allocation so that it works for all chip generations.

The CMT is always on SPI2_HOST (which is the new, consistent naming), and in NRF the selection is now fixed in a way that always selects the second exposed SPI port. (for ESP32 that is VSPI, for S2/S3/C3 that is the HSPI, although that naming is deprecated for these generations anyway) This would be equivalent to using SPI3_HOST, but unfortunately NRF uses the 'old' driver Framework based in SPIClass where these newer consistent defines do not fit.

Lastly, compile options for the new openDTUFusion board are provided and a Device Profile JSON.


tbnobody:
I rebased the commits as they where wide apart.

markusdd:
thanks haha, too much action on master^^

PR #859, opened 2023-04-29
Added links for EBYTE E49-900M20S

OLFDB:
Added links to sources for antenna, module and datasheet of EBYTE E49-900M20S.


tbnobody:
I would close this PR without merging because I added this information to the wiki:
https://github.com/tbnobody/OpenDTU/wiki/CMT2300A-Module

PR #901, opened 2023-05-08
CMT_mit_Display.json

wuscheltom:


tbnobody:
Istead of modifing a existing profile I reworked the whole file and added additional profiles for the display. see abef45524938f359d134f3932980e92724099dd1

PR #1003, opened 2023-06-05
use FSPI for esp32-c3

kiffie:
SPI on the esp32-c3 did not work for me. With this change it does. This should also fix the problem described on [discussion #735](https://github.com/tbnobody/OpenDTU/discussions/735).

I hope this is the right solution. The SPI names and numbering in the HAL libs is confusing.

Maybe, you would also like to accept the additional section in `platformio.ini`.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -60,11 +60,16 @@ build_flags = ${env.build_flags}
 
 
 [env:generic_esp32c3]
-board = esp32dev
-board_build.mcu = esp32c3
+board = esp32-c3-devkitc-02
 custom_patches = esp32c3
 build_flags = ${env.build_flags}
 
+[env:generic_esp32c3_usb]
+board = esp32-c3-devkitc-02
+custom_patches = esp32c3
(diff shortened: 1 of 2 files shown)


tbnobody:
It would be nice to have a generic device  profile without any pre-defined pin assignments and without any board name. The platformio.ini does already contain a profile called `generic_esp32c3`. This could be adjusted if needed. 
I do not own a ESP32-C3 based board, but could you maybe test this profile and your chages regarding the SPI interface if it's working? (you have to upload a device profile or course)

kiffie:
Whats about two profiles: `generic_esp32c3` and `generic_esp32c3_usb`? The latter profile maps the serial port to the USB CDC. The pin assignments are removed.

I successfully tried both profiles on real hardware connected via SPI to an RF24L01+ chip.

tbnobody:
Could you maybe squash these commits into one and force push? (I could do it also, but then you are no more the main author of the patch)

PR #1004, opened 2023-06-05
added config switch to use inverter name as sub topic

### LINKED ISSUE #813: [Request] Configurable MQTT topic

_No response_

It could be beneficial to have a dedicated MQTT topic for each inverter instead of using the serial number. Please consider the case when an inverter needs to be replaced, all subsequent systems would need to be adaped to listen to the new serial number. 

A solution could be to add a new configuration field for each inverter to specify the topic to be used. In the case the field is empty, the serial number shall be used, also to not break already existing implementations

### Additional context

_No response_

Brunhilde:
Added config switch to change sub topic to configured inverter name instead of inverter serial number.

Resolves #813


src/MqttHandleInverter.cpp
```diff
@@ -247,4 +253,17 @@ void MqttHandleInverterClass::onMqttMessage(const espMqttClientTypes::MessagePro
...
         }
     }
+}
+
+String MqttHandleInverterClass::getSubTopic(std::shared_ptr<InverterAbstract> inv)
+{
+    String subtopic;
+
+    if (MqttSettings.getSubTopicInverter()) {
+        subtopic = inv->name();
```
tbnobody, line 263: I think you must take care of some special characters here: `+`, `#` and `*` are wildcard characters and not allowed for publishing.  Also the `$` character is not allowed for mqtt clients. If someone enters a space in the inverter name it would break the home assistent auto discovery. Also slashes (`/`) must not occour in the inverter name. 

So somehow these characters have to be replaced or removed from the topic. I could imagine to replace them by a underscore (`_`)

Brunhilde: Good point. Will add a helper function to replace such characters.

Brunhilde:
@tbnobody 
Just stumbled over the discussion #317. What's the status of this change? Looks like the structure is basically decided. So I'm not sure whether my pull request is still neded...

PR #1059, opened 2023-06-20
add generic s3 build

markusdd:
this is to enable downloadability of a generic esp32s3 build, we might also add the c3 when we are at it.

adapted platformio.ini already, need some guidance on the github flow though.


markusdd:
closing this as @tbnobody has essentially fixed this with https://github.com/tbnobody/OpenDTU/releases/tag/v23.6.21

PR #1063, opened 2023-06-22
Added French Translations

jdloic:


tbnobody:
merged in 3df65b3d37c98bdd261236631c953c7b7a1cd0de

Dom1n1c:
why not add some more translations?
I stumbled across this change and I mean, in case there are native speakers its also easy to correct 😄 
[The "JSON Translator"](https://github.com/mololab/json-translator) or https://translate.i18next.com are easy to use and might be utilized to get localized json files.

So given the [de.json](https://raw.githubusercontent.com/tbnobody/OpenDTU/master/webapp/src/locales/de.json)
I would suggest the following for **ES** (Spanish) and **IT** (Italian): (i dont speak those languages, but it might be a good starting point:
[es.json.txt](https://github.com/tbnobody/OpenDTU/files/11925743/es.json.txt)
[it.json.txt](https://github.com/tbnobody/OpenDTU/files/11925744/it.json.txt)
(and of course those two new languages have to be added in de/en/fr aswell:
de.:
"es": "Spanisch",
"it": "Italienisch",

en.:
"es": "Spanish",
"it": "Italian",

fr.:
"es": "Espagnol",
"it": "italien",

It might be worth to consider possible users by a [list of the sunniest countries in Europe](https://www.euronews.com/travel/2023/03/11/weather-can-make-or-break-a-trip-these-are-europes-sunniest-snowiest-and-rainiest-cities)
So from that list we might add: Greece, Portugal, Albania (some speak italian though), etc 👍

tbnobody:
And you are aware of the fact that each language requires ~10-15kb of memory? I mean no problem if everyone buys a custom board with 16mb flash. In this step it would be easy to also include some PSRam. that would make things even easier in terms of programming....

Dom1n1c:
Sure, most were using 4M Boards. It's quite heavy in terms of the included strings. I know it's a German based Project, but Spain/Italy should be covered. Sure, there are more countries then those mentioned, and according to destatis we were already covering the Top-3 of Western Europe.
Spain and Italy is only a suggestion, but there might be some more. For the Top 6 it might also be an option to include Russian.
Source: https://en.m.wikipedia.org/wiki/List_of_European_languages_by_number_of_speakers

PR #1077, opened 2023-06-25
Rename platformio_override.ini to platformio_override_sample.ini

schlimmchen:
this file is supposed to be changed by users, but users' changes shall never be included in the upstream project. also, it is cumbersome that git marks the file as changed and that it has to be cleaned up when switching branched etc.

rename the file to platformio_override_sample.ini. this way users who actually want to use the platformio_override.ini have something to start from. they only need to copy the sample file. using '_sample' within the file name rather than manipulating the file ending is good for windows users who might not be displaying file extensions in their file explorer.


gregorwolf:
Beside providing the '_sample' I would suggest adjusting the README.md so the documentation is in sync with the files.

schlimmchen:
Good point. Unfortunately, I was having a hard time to incorporate this information without changing the structure of the README.  That is because the `platformio_override.ini` file was mentioned in multiple locations. So... I may have gone a little too far, but I hope that the proposed changes are an overall improvement to the README.

.gitignore
```diff
@@ -8,3 +8,4 @@ platformio-device-monitor*.log
 logs/device-monitor*.log
 platformio_override.ini
 .DS_Store
+platformio_override.ini
```
tbnobody, line 11: Why do you add the file a second time?

schlimmchen: I did not expect the file to be mentioned in `.gitignore` already, I put it in there withouth checking whether or not it was already in there... Thanks for noticing!

schlimmchen:
Rebased once more onto the current master, in case @tbnobody wants to consider this change. Adjusted the respective commit message and the title of this PR to reflect what they actually achieve.

Renaming the `platformio_override.ini` file solves the problem that changes to that file will be tracked regardless of the fact that the file is mentioned in `.gitignore`. I assume that was the intention when adding the file to `.gitignore` originally.

stefan123t:
@schlimmchen do we need to adapt this PR to make the changes to the new documentation project or is the gist of your proposal already adressed by the mention of `platformio_override.ini` in `.gitignore` ?

schlimmchen:
The documentation changes are completely obsolete. Not only because the README.md is basically gone/moved, but also because the changes themselves are kind of outdated.

Concerning the renaming of the `platformio_override.ini` file: No, the gist of this PR is not implemented by adding the file to `.gitignore`. That is because changes are still tracked by Git, as the file was added to the repository at some point. Once tracking of a file started, it will be tracked regardless of `.gitignore`.

However, given that users shall not use `platformio_override.ini` for anything and it seems they actually don't (I have not seen a change to the file in any PR that would annoy anyone), I don't care anymore. The idea is still valid, but it is simply not important. In fact, my updated proposal would be to remove `platformio_override.ini` from the repo an merely mention in the docs that overrides go there if anyone needs them, in which case the file would indeed not be tracked in the future.

I guess the main reason is that most people do not compile from the CLI but from vscode, so they really don't need `platformio_override.ini` at all.

PR #1107, opened 2023-07-02
Open dtu with ePaper 1.54inch lib

dAjaY85:
This is the change, where the ePaper is included as an SW-SPI.
Device Profiles were changed, too.

There is a new Variable UpdatePeriod, this is to control the cycle time of the Displays. This Variable is prepared, you will need to create a WEB UI integration of this Variable.


dAjaY85:
Hi, there are some changes on the lib, to be able to use SW-SPI. The Creator of the original LIB doesn't support SW-SPI officially, which we need to be able to use the other SPI stuff. 

It's possible to minimize the lib, to use only the 1.54" ePaper, if it will help.

tbnobody:
Several patches could also be applied using the patch method which I implemented as `PRE` action in the `platformio.ini` file.

dAjaY85:
I didn't know about a patch method, I saw it only today at Ahoy.

PR #1165, opened 2023-07-23
Verbosity switches

schlimmchen:
These changes implement two user-controlled switches that allow to decrease verbosity of the Hoymiles lib and the MQTT component. The behavior does not change with the new default: all messages are printed as before. But there is now an option to decrease the amount of messages produced by the Hoymiles library and the MQTT component. The former still prints hex dumps of messages in particular, and I argue that the project is past analyzing the protocol in such detail. The latter prints a message for every update of every topic, which can be very noisy.


schlimmchen:
This PR requires some work for it could be able to be merged into the current state of branch master. @tbnobody please let me know if this change is interesting to you. If not, please reject this PR and close it.

PR #1175, opened 2023-07-27
Integration WS2812 with ESP32s3 Board

Tifaifai:
Hello,
I have a ESP32s3 Board with RGB Led on pin 48. This PR use this RGB Led for indicate Network and Inverter Status. (Replace two leds and Blink network status and Inverter Status).

Color status : 
  RED: Network is Off
  GREEN: Network is On
  VIOLET: Network is Connected

  BLUE: Inverter is Off
  YELLOW: Inverter is On
  ORANGE: Inverter is Blink

Coding style of OpenDTU ;)

### DIFF
--- a/include/Led_Strip.h
+++ b/include/Led_Strip.h
@@ -1,0 +1,274 @@
+/*  ---------------------------------------------------------------------------
+    File: Led_Strip.h
+    Author(s):  Allen Kempe <allenck@windstream.net>
+    Date Created: 05/14/2019
+    Last modified: 05/14/2019
+    Description: 
+    This library can drive led strips through the RMT module on the ESP32.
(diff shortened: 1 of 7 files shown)


CommanderRedYT:
Will this also be map-able inside of the web interface?

stefan123t:
@Tifaifai this PR adds a lot of details which are not really necessary to turn a single pixel WS2812 led in a certain color on/off/blinking.

Actually it contains all the example code to make a complete strip of WS2812 display complex patterns, plus the gamma tables for picture perfect brightness & color balance.

Would you want to reduce your PR to the bare minimum to make this PR at least considered for merging ?

CommanderRedYT:
This is just adding a library to the project. Maybe this library already exists in platformio, so then we can add it through this and reduce the code to control the one led.

PR #1188, opened 2023-08-01
Turn LEDs and display off before rebooting.

StefanOberhumer:
If the selected device profile is changed (eg from "LEDs, Display" to "No Output") LEDs and display keeps powered on.
This can also happen if you load a new config.
To prevent this, LEDs and display get powered off before any reboot.

Großen Danke für das tolle Projekt und die ganze Arbeit!  // Many thanks for this great project and the work you did!


tbnobody:
had to implement it a little bit different to make it thread safe. But the idea was great!

It's implemented in c91bd42a770aef2a859b3709c02e0d19dcd444f0 and 701c4906548f50b4168efbbef78a479c5b353809

PR #1189, opened 2023-08-01
user selectable LED brightness setting

moritzlerch:
I was a little annoyed by the bright lights of my OpenDTU especially during night. Instead of changing the hardware, e.g. by adding bigger resistors for the LEDs, I implemented a brightness control for led0 and led1.

Here is a short preview of the brightness control from a users perspective:

https://github.com/tbnobody/OpenDTU/assets/59474941/d438c145-0d9b-4451-9539-ad404c87f04c

 I hope you'll like it ☺️


moritzlerch:
Habe den Build entfernt, ist ja absolut nicht sinnvoll, den hier drinnen zu haben.

stefan123t:
@moritzlerch nice feature, I think your implementation looks sound. Thanks for the PR. Though i dunno if @tbnobody would accept this as he usually tends not to add too many features to the OpenDTU, especially in case it adds additional overhead besides the already existing NRF, CMT Radio and MQTT & REST API code. This is already posing quite some difficulties with timing.

But given the your initial intent (being "annoyed by the bright lights of my OpenDTU especially during night") wouldn't it be better to simply turn off the lights during night. That way you may even use the switches which are already present under Settings > Inverter > Edit Inverter: [x] Poll inverter data at night and [x] Send commands at night ?

@tbnobody maybe you could share some of your thoughts on this PR as feedback for Moritz ?

StefanOberhumer:
@stefan123t I don't think it that code costs a lot of extra time. But @moritzlerch : maybe you can change 
```_ledBrightnessSetting[ledIndex] = brightness * 2.55f;```
to
```_ledBrightnessSetting[ledIndex] = (brightness * 255) / 100;```
in setBrightness() function as I've read float computation is very very time expensive on the ESP32.  (Maybe the compiler already optimize that)
FYI: Integer computation usually is always much faster than float computation on nearly all chips - even divisions!

moritzlerch:
Hey @StefanOberhumer thanks for tuning in! I totally get your point but this method is actually only called once if the user updates the brightness level setting which should not occur that often. The method to update the LED during normal operation (so if the user doesn't change any settings) is actually not dealing with heavy calculations which might cause problems with other resource heavy parts of the code. 

Actually in https://github.com/tbnobody/OpenDTU/blob/3f1b6be8578fa1bb460065eea576c5086204f6c7/src/Display_Graphic.cpp#L203 the same calculation is used to set the contrast of the display. 

I don't know if this would be worth changing if the method isn't used on a regular basis. What do you think?

stefan123t:
@moritzlerch ich weiß nicht ob der PR mit der Brightness so Sinn macht (es sind bei jedem loop die o.g. Berechnungen anzustellen, ob die LED jetzt an / aus zuschalten ist), wie geschrieben halte ich eine komplette Nachtabschaltung der LED für weit sinnvoller und einfacher. Wenn ich Nachts schlafen will, dann brauche ich auch keine gedimmte LED :smiley:

StefanOberhumer:
> @moritzlerch ich weiß nicht ob der PR mit der Brightness so Sinn macht (es sind bei jedem loop die o.g. Berechnungen anzustellen, ob die LED jetzt an / aus zuschalten ist), wie geschrieben halte ich eine komplette Nachtabschaltung der LED für weit sinnvoller und einfacher. Wenn ich Nachts schlafen will, dann brauche ich auch keine gedimmte LED 😃

I like it!
I want to have an idea if my opendtu device is still connected to the WIFI - even if I come home late at night and the display is already in screen-saver-mode - so the "night mode" is not an valid option.
I think the change(s) in this PR does not waste much processing time nor a lot of memory!

tbnobody:
Implemented in 9ae791edd423e0b9e379ed60de04f24b34ee215c

Thanks for the idea.

PR #1211, opened 2023-08-05
Added links to another case and PCB

foorschtbar:
This case and PCB differs from the others by his compactness and mounting options on a wall. I would be happy if the PR is accepted.


T3rm1:
What are the notable changes of the breakout board compared to the original (which is already linked here)?

foorschtbar:
The original PCB had some restrictions, which i had fixed in my fork:
- Cap under the Transceiver 
- Only compatible with short ESP32 DevBoards
- No sliding rails for my (smaller) case

tbnobody:
The links are added to the documentation.

PR #1270, opened 2023-08-23
Fix: Home Assistant autoconf change inverter limits

madmartin:
The announced limits for inverters in HA autoconfig messages are not suitable for the sub-project OpenDTU-onBattery. Error messages found in HA logs are

Invalid value for number.limit_persistent_relative: 0.1 (range 1.0 - 100.0)
Invalid value for number.limit_nonpersistent_absolute: 0.6 (range 10.0 - 2250.0)
Invalid value for number.limit_persistent_absolute: 0.6 (range 10.0 - 2250.0)
Invalid value for number.limit_nonpersistent_absolute: 9.6 (range 10.0 - 2250.0)
Invalid value for number.limit_persistent_absolute: 9.6 (range 10.0 - 2250.0)

This fix changes the limit minimum to 0 (zero) instead of 1 or 10.

### DIFF
--- a/include/MqttHandleHass.h
+++ b/include/MqttHandleHass.h
@@ -59,7 +59,7 @@ private:
     void publish(const String& subtopic, const String& payload);
     void publishField(std::shared_ptr<InverterAbstract> inv, ChannelType_t type, ChannelNum_t channel, byteAssign_fieldDeviceClass_t fieldType, bool clear = false);
     void publishInverterButton(std::shared_ptr<InverterAbstract> inv, const char* caption, const char* icon, const char* category, const char* deviceClass, const char* subTopic, const char* payload);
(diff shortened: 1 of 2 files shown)


madmartin:
Please also see https://github.com/helgeerbe/OpenDTU-OnBattery/issues/396

tbnobody:
What happens to the inverter if the limit is set to zero? Is it still reachable and just don't produce any output?

madmartin:
> What happens to the inverter if the limit is set to zero? Is it still reachable and just don't produce any output?

Yes exactly that.

tbnobody:
Implemented in b4ead9d6ec42609ed8fe7ca32d815ad4b29f32ca

There where some other locations which needed to be changed as well

Issue #1325, opened 2023-09-11
Dynamic Limit API


unaiur:
### Is your feature request related to a problem? Please describe.

I want to implement dynamic power limits using Web API of OpenDTU to ensure we do not dump more than 600w to the network as required in most countries in EU.

The problem is that, if the WIFI or the controller dies, the limit may be left unmanaged at a value that is too high.

### Describe the solution you'd like

The easiest solution I can envision is a new dynamic limit WebAPI that receives a temporal limit in Watts and a timeout in seconds. This API will cancel any previous scheduled reset that may be pending, apply the new temporal limit as non-persistant and schedule a limit reset in the given amount of seconds.

This way, if no new limit is received in X seconds, we revert to the persistent limit. This will reduce the chances of uncompliant energy dumps greatly.

unaiur:
If there is no easy way to retrieve the persistent limit, I would add an extra parameter to the new API with the fallback limit to apply when the timer expires.

Heiko-L:
> I want to implement dynamic power limits using Web API of OpenDTU to ensure we do not dump more than 600w to the network as required in most countries in EU.

I think you don't understand the intention to the (8/6)00W limit.

Why not set the persistent limit to 600W and change only non persistent afterwards?

unaiur:
I want to implement some logic that every second measures the power consumed in my home and limits the panels output to the consumed power + 400W.

Suppose that I turn on my 900W microwave; the limit is raised to 1300W but some seconds later my wifi hangs and no updates are sent anymore. One minute later my coffee is hot, the microwave stops and I am dumping 1300W into the electricity network, way above the 600W limit, and it will continue doing so while the weather is sunny and the WiFi is down.

I would like to reduce the likelihood of that scenario.

unaiur:
Of course, the persistent limit is set at 600W. I fully agree on that and now I have permanently set at this limit. 

But, when my home consumption is above 600W and no power is dumped on the grid, I would like to raise the limit to my current consumption. 

For example, if I turn on the oven (which consumes ~2000W), I would like to temporarily raise the limit to 1200 (maximum for my panels). If I turn on the microwave (~900W), I would like to raise the limit to 900W.

No problem here. I have an energy meter that can provide measures multiple times per second and can raise the limit when possible.

The problem is reducing the limit when the oven or microwave finishes.

If everything goes well, the limit will be readjusted in milliseconds and there is no problem at all.

[...]

PR #1355, opened 2023-09-18
Safe Limits API (#1325)

unaiur:
When an inverter has a safe limit timeout enabled, a new /api/limit/safe endpoint is enabled to enact limits safely.

The caller needs to ensure that the limit is confirmed before the timeout expires; otherwise, the limit is rolled back to the safe limit value.


RobertLukan:
This is a very good feature. Do you think it will get merged ?

stefan123t:
@unaiur maybe it would be a good idea to add this PR to OpenDTU-OnBattery as it would be more fitting there ?

This is similar to a "hard" temporary limit, which will be used as a fallback after N minutes of idle time, 
i.e. no updated temporary limit or otherwise a heartbeat from the external / internal (dynamic) limit controller will send the safe fallback limit to prevent e.g. deep decharge of a battery powered system.
See #1093 for another change to the Limit API making it more robust.

unaiur:
Sorry, I sold my Hoymilles inverter and replaced it with an Anker
PowerBank, so I cannot contribute anymore.

Issue #759, opened 2023-03-28
[Request] Add additional MQTT topics


DejanBukovec:
### Describe the solution you'd like

Please add next MQTT topics:
- Firmware Version / Git Hash
- Firmware Update

That way we can get some notifications on Home Assistant that new update is availible...

DcFamas:
Hey there,
I also want to ask for a sum up of this values:

![grafik](https://user-images.githubusercontent.com/35306967/228768074-84b835b4-b586-4975-a62c-d64e99fa6a20.png)
(Screenshot by AhoyDTU)

Best regards :)

tbnobody:
Please see https://github.com/tbnobody/OpenDTU/discussions/317#discussioncomment-4345649

Currently there are just other prios (like HMS/HMT)

stefan123t:
Here Eventlog Code and Count have been requested as MQTT Topics too: #424

PR #1368, opened 2023-09-22
feat(home assistant): add connection status entity for inverter

CFenner:
This adds a connection status entity in the *diagnostic* section of the inverter. I have these kind of sensors on my ESPhome devices and find them very handy.

Bildschirmfoto 2023-09-22 um 13 39 59

It's basically the same as `reachable` with a device class, but I anyhow created a new one as I do not know if renaming of the entity would be fine for the maintainers.


stefan123t:
@CFenner so renaming `reachable` to `Status` would do the trick ?
Is it also possible to use lowercase `status` ?

This breaking change should be considered as part of #759.

CFenner:
puh, hardly can remember what I did here 😄 ...

> Is it also possible to use lowercase status ?

Typically HA has capital first letter.

> Or does it simply need to be defined as connectivity, diagnostic ?

Yes, with the device class `connectivity`.

CFenner:
Looks like this is already implemented.. reachability now has a device class and category. I will close this, as I'm also fine with that name.

Issue #1428, opened 2023-10-14
[Request] mDNS / DNS-SD / Bonjour to announce opendtu-<hash>.local Name


stefan123t:
### Is your feature request related to a problem? Please describe.

Users without a proper DHCP + DNS Server eg Speedport 723V Typ B routers may not be able to figure out the IP assigned to OpenDTU by the routers DHCP after some reboot.

### Describe the solution you'd like

Announce the name opendtu-<hash>.local via mDNS in the local network. That way users can access the DTU more easily.

To be honest I change my name to opendtu or opendtu-fusion in my FritzBox but other / novice users may not have that option in their routers.

### Describe alternatives you've considered

Adding a static mapping to the DHCP Server to assign a static IP and Name to the ESP MAC address. But this feature is not available in all routers. I have flashed freetz over my old and nowadays outdated FritzBox 3xxx just to have a proper DHCP with this functionality. But not all users may be willing / able to flash custom firmware on their routers to workaround this issue ;)

### Additional context

See PR #695 for the necessary changes and the AhoyDTU code where it is implemented too

stefan123t:
By the way get well soon first!

CommanderRedYT:
If you want I can submit a new PR with

- Webapp mDNS Enable/Disable
- Announce the name so opendtu-xxxxx.local will be possible
- Add records that show devices that OpenDTU has a website 
- Custom record so one can scan the network for OpenDTU auto discovery.

Open questions would be:

- Enabled by default?

stefan123t:
@CommanderRedYT the name of your OpenDTU is already (pre)set under **Settings > Network Settings**
Hostname: `OpenDTU-%06X`

Here `%06X` is replaced with the last 3-bytes/6-chars of your ESP32 chips MAC-Address in Hex representation.
I regularly change this to either opendtu or opendtu-fusion to distinguish my two DTU targets, 
though I know other developers do have more than just two DTUs.
I enjoy the simplicity of speaking names without hex chars, which I won't remember easily :older_man:.

The question for me is more how does that work when setting the Hostname in my case, even though mDNS is not used in OpenDTU yet ?

CommanderRedYT:
> The question for me is more how does that work when setting the Hostname in my case, even though mDNS is not used in OpenDTU yet ?

Hostname Solution only works when DNS and DHCP are communicating afaik

CommanderRedYT:
[...]

PR #1441, opened 2023-10-20
Implemented mDNS

CommanderRedYT:
This PR implements mDNS for OpenDTU. It adds two services, one for the commonly used _http._tcp service, and one custom one (_opendtu._tcp) for auto-discovery.

It also implements the needed changes for the Web-App.

Flash Statistics:

```sh
$ du .pio/build/generic_esp32/firmware.bin

# before: 1512
# after: 1540
```

This PR was inspired by https://github.com/tbnobody/OpenDTU/pull/695.

### DIFF
--- a/include/defaults.h
+++ b/include/defaults.h
@@ -20,6 +20,8 @@
 #define WIFI_PASSWORD ""
 #define WIFI_DHCP true
 
+#define MDNS_ENABLED false
+
 #define NTP_SERVER "pool.ntp.org"
 #define NTP_TIMEZONE "CET-1CEST,M3.5.0,M10.5.0/3"
 #define NTP_TIMEZONEDESCR "Europe/Berlin"
--- a/src/NetworkSettings.cpp
+++ b/src/NetworkSettings.cpp
@@ -8,6 +8,7 @@
 #include "PinMapping.h"
 #include "Utils.h"
 #include "defaults.h"
+#include <ESPmDNS.h>
 #include <ETH.h>
 
 NetworkSettingsClass::NetworkSettingsClass()
@@ -107,6 +108,35 @@ void NetworkSettingsClass::raiseEvent(network_event event)
                 entry.cb(event);
             }
         }
+    }
+}
+
+void NetworkSettingsClass::handleMDNS()
+{
+    bool mdnsEnabled = Configuration.get().Mdns_Enabled;
+
+    if (lastMdnsEnabled == mdnsEnabled) {
+        return;
+    }
+
+    lastMdnsEnabled = mdnsEnabled;
+
+    MDNS.end();
+
+    if (!mdnsEnabled) {
+        return;
+    }
+
+    if (MDNS.begin(getHostname())) {
+        MessageOutput.print("MDNS responder starting...");
+
+        MDNS.addService("http", "tcp", 80);
+        MDNS.addService("opendtu", "tcp", 80);
+        MDNS.addServiceTxt("opendtu", "tcp", "git_hash", AUTO_GIT_HASH);
+
+        MessageOutput.println("done");
+    } else {
+        MessageOutput.println("Error setting up MDNS responder!");
     }
 }
 
@@ -218,6 +248,8 @@ void NetworkSettingsClass::loop()
     if (dnsServerStatus) {
         dnsServer->processNextRequest();
     }
+
+    handleMDNS();
 }
 
 void NetworkSettingsClass::applyConfig()
--- a/include/Configuration.h
+++ b/include/Configuration.h
@@ -66,6 +66,8 @@ struct CONFIG_T {
     char WiFi_Hostname[WIFI_MAX_HOSTNAME_STRLEN + 1];
     uint32_t WiFi_ApTimeout;
 
+    bool Mdns_Enabled;
+
     char Ntp_Server[NTP_MAX_SERVER_STRLEN + 1];
     char Ntp_Timezone[NTP_MAX_TIMEZONE_STRLEN + 1];
     char Ntp_TimezoneDescr[NTP_MAX_TIMEZONEDESCR_STRLEN + 1];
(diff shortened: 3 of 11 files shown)


CommanderRedYT:
Note: I have commented out the deinit code for mDNS inside NetworkSettings.cpp. With it commented in, the ESP crashes inside lwip tcp functions. I am unfortunately not able to debug this in detail, so help is required there. Other option would be to maybe add a note that after deactivating mDNS, user has to restart the ESP.

CommanderRedYT:
I cleaned it up a little bit more. I looked at some other projects that use esp32 and mDNS (like WLED for example), and they all only enable/disable mDNS on boot. This means that every time you'd change this option you would have to reboot.

I never had problems with it when turning it on, so if there are problems I could rewrite it so that enabling will work, but disabling only works on boot. But imo it works fine like this.

include/defaults.h
```diff
@@ -20,6 +20,8 @@
 #define WIFI_PASSWORD ""
 #define WIFI_DHCP true
 
+#define MDNS_ENABLED true
```
tbnobody, line 23: Please set false as default to not break current default functionality

webapp_dist/index.html.gz
tbnobody, line 1: Please remove .gz file from PR (I will compile the webapp on final merge)

webapp_dist/js/app.js.gz
tbnobody, line 1: Please remove .gz file from PR (I will compile the webapp on final merge)

webapp_dist/zones.json.gz
tbnobody, line 1: Please remove .gz file from PR (I will compile the webapp on final merge)

src/NetworkSettings.cpp
```diff
@@ -9,6 +9,7 @@
 #include "Utils.h"
 #include "defaults.h"
 #include <ETH.h>
+#include <ESPmDNS.h>
```
tbnobody, line 12: Please sort includes in alphabetic order (ESPmDNS before ETH)

src/NetworkSettings.cpp
```diff
@@ -110,6 +111,31 @@ void NetworkSettingsClass::raiseEvent(network_event event)
     }
 }
 
+void NetworkSettingsClass::handleMDNS()
+{
+    bool mdnsEnabled = Configuration.get().Mdns_Enabled;
+
+    if (lastMdnsEnabled != mdnsEnabled) {
```
tbnobody, line 118: Please remove the lot of nested if statements (e.g. by doing the following)

```cpp
bool mdnsEnabled = Configuration.get().Mdns_Enabled;

if (lastMdnsEnabled == mdnsEnabled) {
    return;
}
lastMdnsEnabled = mdnsEnabled;

MDNS.end();

if (!mdnsEnabled) {
    return;
}

if (MDNS.begin(getHostname().c_str())) {
```

src/NetworkSettings.cpp
```diff
@@ -110,6 +111,31 @@ void NetworkSettingsClass::raiseEvent(network_event event)
...
+        MDNS.end();
+
+        if (mdnsEnabled) {
+            if (MDNS.begin(getHostname().c_str())) {
```
tbnobody, line 122: There is no need to call `.c_str()` as the begin method uses String internaly anyways.

tbnobody (changes requested):
Would you mind combining the commits into one with git squash?

CommanderRedYT:
@tbnobody Thanks for the Review, I have resolved all of the issues :)

src/NetworkSettings.cpp
```diff
@@ -110,6 +111,33 @@ void NetworkSettingsClass::raiseEvent(network_event event)
     }
 }
 
+void NetworkSettingsClass::handleMDNS()
+{
+    bool mdnsEnabled = Configuration.get().Mdns_Enabled;
+
+    if (lastMdnsEnabled == mdnsEnabled)
```
tbnobody, line 118: Please add curly braces even if the statement contains only one line.

src/NetworkSettings.cpp
```diff
@@ -110,6 +111,33 @@ void NetworkSettingsClass::raiseEvent(network_event event)
...
+        return;
+
+    lastMdnsEnabled = mdnsEnabled;
+
+    MDNS.end();
+
+    if (!mdnsEnabled)
```
tbnobody, line 125: Please add curly braces even if the statement contains only one line.

src/NetworkSettings.cpp
```diff
@@ -2,12 +2,13 @@
 /*
  * Copyright (C) 2022 Thomas Basler and others
  */
-#include "NetworkSettings.h"
 #include "Configuration.h"
 #include "MessageOutput.h"
+#include "NetworkSettings.h"
```
tbnobody, line 7: The header file belonging to the cpp file should stay in the first line (this is also done by the auto formatter of vscode)

tbnobody (changes requested):
Sorry, just had to add some remarks (it triggered my "Internal Monk")

src/NetworkSettings.cpp
CommanderRedYT: Normally I would have a empty line between .h file of a .cpp file and all other headers :P

CommanderRedYT:
Okay, fixed everything again :+1:

CommanderRedYT:
Btw we should maybe link issue #1428 with this PR

CommanderRedYT:
@tbnobody Any updates?

tbnobody:
It's already included in my local branch for several days. Just had no time for a release yet. Will (hopefully) publish it tomorrow or tuesday.

CommanderRedYT:
> It's already included in my local branch for several days. Just had no time for a release yet. Will (hopefully) publish it tomorrow or tuesday.

Awesome, thank you very much! Have a nice evening

PR #1466, opened 2023-10-31
webapp: trim username

### LINKED ISSUE #1456: [Request] Trim whitespaces before and after login username

Mobile autocompletion usually puts a whitespace after the completed word. 
When that happens, it is hard to spot, and the login name becomes `admin ` instead of `admin`. 
So the connection is refused.

Trim login names

- Either trim whisteapces
- Either prevent autocompletion from being allowed for the username field

### Additional context

_No response_

flooxo:
Resolves #1456

This change trims the leading and trailing whitespaces of the username in case it gets auto filled.


tbnobody:
Implemented in 6a64fe6e1a9a5c59093111c6b94a45d746b6dc13

PR #1505, opened 2023-11-16
Esp32 eth poe ecrow

khancyr:
Esp32 stick eth poe board support https://www.elecrow.com/esp32-stick-poe-a.html
Fix LilyGO_T_ETH_POE ETH pins

### DIFF
--- a/docs/DeviceProfiles/esp32_stick_poe_a.json
+++ b/docs/DeviceProfiles/esp32_stick_poe_a.json
@@ -1,0 +1,22 @@
+[
+    {
+        "name": "Esp32-Stick-PoE-A",
+        "nrf24": {
+            "miso": 2,
+            "mosi": 15,
+            "clk": 14,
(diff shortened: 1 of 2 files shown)


tbnobody:
Please don't use this custom profiles. Instead create a device profile json. Several samples can be found here: <https://github.com/tbnobody/OpenDTU/tree/master/docs/DeviceProfiles>

khancyr:
ok I will create a new profile for the esp32_stick_poe
For the Lily poe, the config is currently broken, do you prefer to keep the fix, or full removal as we got the custom profile ?

tbnobody:
If it's broken then I would prefer removing it from the `platform.ini` and have a proper device profile instead.

khancyr:
updated with profile.
One bad things on profiles is that we lost the comments line to find the board official docs. If we can have another field for this that would be great.

PR #1586, opened 2023-12-23
http response header tag 'ETag' needs quotes.

StefanOberhumer:
As per RFC7232 ( https://www.rfc-editor.org/rfc/rfc7232#section-2.3 ) and MDN docs ( https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#directives ) the value of the ETag in the http header response has to be between quotes.


StefanOberhumer:
@stefan123t :
Thx for the praise.
I still have some issues things getting not refreshed sometimes and still investigating that. 
That was the reason I stumbled over the ETag syntax - which did not solved my issue but seemed to be implemented incorrectly.
I'm happy the fix helped in your case!
Hopefully I find the reasons why my browser (still) does not refresh sometimes!

PR #1620, opened 2024-01-04
WebApp option to disable tiny Diagram

Henrik-Ingenieur:
I've added an option to the WebApp to disable the diagram on the display. This was discussed and desired by some users in discussion [#1504](https://github.com/tbnobody/OpenDTU/discussions/1504)


tbnobody:
Had to implement it a little bit different because I wanted a fullscreen diagram as well. So I introduced a display mode instead of a simple on/off button.

Lineflyer:
Please allow a question:
I just upgraded to [v24.1.18](https://github.com/tbnobody/OpenDTU/releases/tag/v24.1.18) and wanted to try out the new diagram options.
I do however not see them in Hardware/Display, where they should be if I interpret the related PR correctly.

Thats how it looks to me:
image

Rebooted and powercycled but no change.
Any advice?

tbnobody:
Reload the website or clear your browsers cache...

Lineflyer:
ah damn... I should memorize this. We had this on another occasion.  It worked :)

PR #1639, opened 2024-01-13
refactor webapp/utils/time

nkappler:
This change refactors the `timestampToString` function for the webapp by replacing a bunch of math with two calls to standard JavaScript APIs.

Doing this comes with a bunch of improvements:

1. implementation size reduced by ~50%
2. included translation and correct pluralization for every language for free:

|Lang|Before|After|
|-|-|-|
| EN | 0 days hh:mm:ss | 0 days ago hh:mm:ss |
| EN | 1 days hh:mm:ss | 1 day ago hh:mm:ss |
| EN | 2 days hh:mm:ss | 2 days ago hh:mm:ss |
| DE | 0 days hh:mm:ss | vor 0 Tagen hh:mm:ss |
| DE | 1 days hh:mm:ss | vor 1 Tag hh:mm:ss |
| DE | 2 days hh:mm:ss | vor 2 Tagen hh:mm:ss |
| FR | 0 days hh:mm:ss | il y a 0 jour hh:mm:ss |
| FR | 1 days hh:mm:ss | il y a 1 jour hh:mm:ss |
| FR | 2 days hh:mm:ss | il y a 2 jours hh:mm:ss |

### DIFF
--- a/webapp/src/utils/time.ts
+++ b/webapp/src/utils/time.ts
@@ -1,17 +1,8 @@
-export const timestampToString = (value: number, includeDays = false): string => {
-    const days = Math.floor(value / (24 * 60 * 60));
-    const secAfterDays = value - days * (24 * 60 * 60);
-    const hours = Math.floor(secAfterDays / (60 * 60));
-    const secAfterHours = secAfterDays - hours * (60 * 60);
-    const minutes = Math.floor(secAfterHours / 60);
-    const seconds = secAfterHours - minutes * 60;
+export const timestampToString = (timestampSeconds: number, includeDays = false): string => {
+    const timeString = new Date(timestampSeconds * 1000).toLocaleTimeString([], { timeZone: "UTC" });
+    if (!includeDays) return timeString;
 
-    const dHours = hours > 9 ? hours : "0" + hours;
-    const dMins = minutes > 9 ? minutes : "0" + minutes;
-    const dSecs = seconds > 9 ? seconds : "0" + seconds;
-
-    if (includeDays) {
-        return days + " days " + dHours + ":" + dMins + ":" + dSecs;
-    }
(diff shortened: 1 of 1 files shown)


tbnobody:
It seems that I have to undo this PR. Please see #1649

Also the uptime now looks a little bit weird.
image
or with positive days
image


I mean regarding the issue mentioned above the fix would be to pass `this.$i18n.locale` to the `Intl.RelativeTimeFormat` object but specifing the uptime in AM and PM looks wrong for me. I would have expected something like `1 day 16:34:43`.

tbnobody:
I could imagine that something like this would do the job:

```ts
export const timestampToString = (timestampSeconds: number, locale: string, includeDays = false): string => {
    const timeString = new Date(timestampSeconds * 1000).toLocaleTimeString([], { timeZone: "UTC", hour12: false });
    if (!includeDays) return timeString;

    const secondsPerDay = 60 * 60 * 24;
    const days = Math.floor(timestampSeconds / secondsPerDay);
    return new Intl.RelativeTimeFormat(locale).format(-days, "day") + " " + timeString;
}
```

nkappler:
I've also noticed this with the new version.

The language is easily fixed by passing a locale as the first parameter, eg. "en" or "de".
Currently it uses the browser default which could be different from the webapp setting.

For the uptime we could trim the leading  word as the TimeRange API is not very cobfigurable as it seems

Grothesk242:
On my system:

Betriebszeit | 1 day ago 2:38:19 AM

Browser language is set to US English. Yet I would like to see the german  time format.

nkappler:
I've created a fix  (#1652) using vue_i18n for the day localization and pluralization and fixed the time to 24 hour format.

I've also passed the locale to the time function, which currently wouldn't be necessary since all three languages use the same hour formatting, but if there are ever new locales, it might be useful.

PR #1652, opened 2024-01-15
fix #1649

### LINKED ISSUE #1649: time.ts BUG
### What happened?

In new time.ts version is bug. It use local language for days even if in DTU is set English...
Commit:
https://github.com/tbnobody/OpenDTU/commit/6166ecdb7d2216423165cda08fba971a60f0c261

New version show:
slika

Old version show:
slika

Language is in both cases English, because Slovenian is not supported...

### To Reproduce Bug

Update to laters version...

### Expected Behavior

Show "days" depend of selected language...

### Install Method

Pre-Compiled binary from GitHub

4741727

### Relevant log/trace output

_No response_

### Anything else?

_No response_

nkappler:
one of the goals of my pull request, besides simplifying the code was to have localization.

It's nice that the browser can handle this, but for consistency, we'll go with vue-i18n since it is already available

### DIFF
--- a/webapp/src/components/EventLog.vue
+++ b/webapp/src/components/EventLog.vue
@@ -31,7 +31,7 @@ export default defineComponent({
     computed: {
         timeInHours() {
             return (value: number) => {
-                return timestampToString(value);
+                return timestampToString(this.$i18n.locale, value)[0];
             };
         },
     },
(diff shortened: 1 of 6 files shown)


nkappler:
don't merge yet, I've just found a bug with the translation variant selection

tbnobody:
Merged manually

PR #1655, opened 2024-01-16
Ux/UI improvements

nkappler:
On my Samsung S22 Ultra which has one of the largest phone screens and with medium small text size, the home page has lots of redundant spacing, which leads to an unnecessary scrollbar in the inverter channel info tables.
Another reason for this is the table header, which - especially in German - makes the unit column much wider than necessary. As you can see the table is cut off having a scrollbar, even though the table content fits without a problem:

grafik

<hr />

 #### This PR makes a few slight changes to use the space more wisely, along with a few other improvements:

Every View had a redundant main container with `class="container-fluid"`, causing unnecessary 1rem margin. This PR removes it:
grafik

<hr />

The Inverter Channel Info Table Header `Property | Value | Unit` caused layout shifting while (arguably) adding next to no value. This PR removes it, along with the table margin:

|Before|After|
|-|-|
| grafik | grafik |

As you can see there's now a lot more available space in the table, meaning that scrollbar folding will only appear on very very small screens.

<hr />

Vertically Centering the OpenDTU Header Text and Icon

|Before|After|
|-|-|
| grafik | grafik |

### DIFF
--- a/webapp/src/App.vue
+++ b/webapp/src/App.vue
@@ -1,8 +1,6 @@
 <template>
     <NavBar />
-    <main class="container-fluid">
-        <router-view />
-    </main>
+    <router-view />
 </template>
 
 <script lang="ts">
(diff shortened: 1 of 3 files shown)


nkappler:
> [!CAUTION]
> I've messed up my branch maintenance and this PR also includes the changes for PR #1652 
> If you don't want to merge both, let me know and I'll crumble apart the commits...

tbnobody:
merged manually

PR #1656, opened 2024-01-16
Resolves recently introduced inability of HTTPPowerMeter to use non-standard ports 

Fribur:
should resolve problems reported in issue https://github.com/helgeerbe/OpenDTU-OnBattery/issues/611 related to utilizing non-standard ports.


Fribur:
pull request accidently created against upstream of OpenDTU-OnBattery

PR #1675, opened 2024-01-21
Add Spanish locale to display

foroae:
Add Spanish locale to display


schneeer:
Not interesting to me, but Spanish is also widely spoken.

I think something would happen here if you had taken the trouble to translate the app texts too.

tbnobody:
Spanish language is implemented as a language pack

PR #1679, opened 2024-01-23
Add Spanish locale to the web

foroae:
Add Spanish locale to the web


3DJupp:
Yeah, one of the more sunnier countries. 🇪🇸 , but also 🇬🇷 , 🇵🇹 might be an option.

tbnobody:
Please see https://github.com/tbnobody/OpenDTU/pull/1830#issuecomment-2051951742

This PR is not forgotten... Just working on another solution

stefan123t:
@tbnobody as this other NLS solution may help tremendously to reduce the code and factor out the common language strings into some shared dictionary / JSON referenced by the Vue app:
Do you already have a working prototype or can we help already by providing these NLS / PO dictionary archives ?

tbnobody:
Spanish language is implemented as a language pack

PR #1703, opened 2024-01-29
Add support for Waveshare 1.54" ePaper

mtavenrath:
This is a minimal invasive version of @dAjaY85's ePaper integration with a few fixes. Until this has been merged it should be way easier to rebase it without the merge conflicts due to the renaming of the DisplayClass.


zoeck-dev:
I hope that the pull request will be integrated into the master branch.

I would like to switch to OpenDTU, but I am already using the ePaper Display with AhoyDTU. I don't really want to buy a new display :)

mtavenrath:
I've updated the PR to master & fixed the issue with the display not updating the last timestamp.

stefan123t:
@mtavenrath could you also check if there is still something missing regarding the merge of #2306 this week ?

mtavenrath:
@stefan123t To my best understanding the ePaper support does not require the SPI manager since it added a SW SPI implementation to the waveshare library. Thus no changed are required for the SPI manager itself.

@tbnobody I tried to rebase the change again today and it's again quite some work due to the changes to the display classes. Before rebasing this again, is there a chance to get this change merged? There is a slight difference in how the two different display classes work which results made @dAjaY85 implement two display classes, one for the ePaper which less frequent updates and one for 'realtime' displays. Everytime the display class interface is changed I have to figure out the changes and apply them to both classes as well.

wusel42:
Quoting [mtavenrath](https://github.com/mtavenrath):

> Before rebasing this again, is there a chance to get this change merged?

It's now three months unanswered, thus the general public should consider this as as a "no" — which is a shame as low-power ePaper displays should be considered the norm for these devices, especially as display updates only happen every 5 to 10 seconds anyway. Please reconsider acceptance of this PR.

PR #1729, opened 2024-02-10
Fix: make efficient use of available display area

schlimmchen:
Fixes calculation of the text baselines, using `getAscent()` in favor of `getMaxCharHeight()`, which includes ascent and descent. This moves the first text up and allows to insert margin between the lines until the display area is fully utilized.
    
On large displays, if the small diagram is selected, keep the first line rather low to avoid collision with the diagram y-axis label. in this mode, there is still more space between the text lines as before, allowing for improved readability.

I always thought that the display was quite crammed, especially the last line. While implementing https://github.com/helgeerbe/OpenDTU-OnBattery/pull/658 I noticed that something was off with the baselines and that we could do better without compromising, since the previous implementation wastes lines mainly on the bottom of the display, also on the top if no diagram is drawn. The y of today nearly touches the l of total.

image

I found that especially the baseline (`_lineOffsets`) for the first line was off. The reason is that `getMaxCharHeight()` returns a value greater than the font's ascent, i.e., it includes the font's descent. The value which should be used is `getAscent()`. The next line's baseline is then advanced by the previous line's descent, making sure that they will not collide.

image
image

The placement is such that even when using the screensaver with the small diagram, the solar power will not collide with the diagram y-axis label.

Using this code
```
    if (!_isLarge) {
        _display->drawBox(84, 0, 1, 49);
        _display->drawBox(0, 48, 84, 1);
    }
    _isLarge = (_mExtra % 40 < 20);
    calcLineHeights();
```
I made sure that the change works for small displays as well.

image

### DIFF
--- a/src/Display_Graphic.cpp
+++ b/src/Display_Graphic.cpp
@@ -67,11 +67,19 @@ void DisplayGraphicClass::init(Scheduler& scheduler, const DisplayType_t type, c
 
 void DisplayGraphicClass::calcLineHeights()
 {
-    uint8_t yOff = 0;
+    bool diagram = (_isLarge && _diagram_mode == DiagramMode_t::Small);
+    // the diagram needs space. we need to keep
+    // away from the y-axis label in particular.
+    uint8_t yOff = (diagram?7:0);
     for (uint8_t i = 0; i < 4; i++) {
         setFont(i);
-        yOff += (_display->getMaxCharHeight());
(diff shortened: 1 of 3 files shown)


tbnobody:
This PR leads to the effect that the text "Offline" is no more centered but left aligned on the display. This doesn't look very nice as the rest of the text is centered.

PR #1730, opened 2024-02-10
Feature: allow to select ISO 8601 date format for display

schlimmchen:
After completing this I realized: Maybe this should be a global option instead and other places where a date is displayed (log? web app?) should also follow the global setting?


StefanOberhumer:
As to use another "fixed" date format:
Wouldn't it be an idea to allow configuring the "format" of the strftime() function?

schlimmchen:
Superseded/completed/solved by Thomas through 70f301941b140cf77533f361f1b09e98069048be :muscle:

PR #1796, opened 2024-02-27
Only show Tooltip if Version is not recent

Sn0w3y:


tbnobody:
Please correct the intends in the SystemInfoView, Squash the commits (as it is one feature) and force push again.

Issue #319, opened 2022-11-08
[Request] OTA Update via URL 


D3R-ST3FAN:
### Is your feature request related to a problem? Please describe.

I have a couple of neighbors that now also run OpenDTU.
They are not into self compiling (they are using the Olimex POE Variant).
I do compile binaries for them and make it available via a URL.

What about OTA Update via URL? Like e.g. tasmota does.

### Describe the solution you'd like

introduce a field to provide an OTA URL and an Update button.

stefan123t:
Tasmota uses safeboot to ensure OTA update via OTA URL does not break things.

See OTA_URL and OtaUrl in 
* https://github.com/arendst/Tasmota/tree/master/tasmota/berry/modules/Partition_Wizard/partition_wizard.be
* https://github.com/arendst/Tasmota/tree/master/tasmota/berry/animate_demo/leds_animation.be

I do not know if we have enough space for app0 and app1 partitions to store two firmware images side-by-side ?

CommanderRedYT:
@stefan123t It is a/b partitions, right? I know that the esp-idf has some flag for marking a partition as good.

When you use this, it is initially marked as bad, so in case something crashes, the bootloader goes back to the old image.

The code can do checks and when it thinks its valid (and maybe after 1 minute) and then mark the partition.

https://docs.espressif.com/projects/esp-idf/en/v5.3.1/esp32/api-reference/system/ota.html#app-rollback

stefan123t:
@CommanderRedYT thanks for adding the ESP-IDF context. I thought that there is two boot environments provided by so called safeboot in Tasmota. But this explains the underlying working principle, Thanks!
For me the question is more how big can we make such an app image. If we consider 4MB of the average ESP32 flash mem, it can be max 2x~1.8MB to leave some slack for the safeboot loader and the basic pin_mapping.json and config.json on the SPIFFS.
At the moment we start to require 8MB ESP32S3 for eg OpenDTU-OnBattery in order to fit all the extra functionality, ie this might work then but nly on a newer ESP32S3 with 16MB for such large app images.
Maybe this ratio gets a bit better if we can source out the NLS strings into some shared dictionary / JSON archive that we can import into the Vue app. See PR #1619 for @tbnobody|s stab into that direction.

CommanderRedYT:
https://github.com/tbnobody/OpenDTU/blob/master/partitions_custom_4mb.csv

I believe this is the partition table. I would propose the following.

We have two options:

a)

[...]

PR #1830, opened 2024-03-12
Added Italian language

CreasolTech:
I just changed 3 files, did not compile and test everything.
If there is any issue, just let me know and I can solve it.
Thanks a lot for the great software.


tbnobody:
If there was no memory limitation I would add as much languages as possible.... I am currently working on a solution that allows a upload of language packs during runtime (compareable to device profiles).

stefan123t:
@tbnobody as this NLS language pack solution may help tremendously to reduce the overall code and factor out the common language strings into some shared dictionary / JSON referenced by the Vue app, this would also allow maybe to introduce a “safeboot” environment which aids in automatically switching between OTA updates, see #319

tbnobody:
With the next version it will (hopefull) possible to upload custom language packs. For the beginning only custom languages for the webapp will be available. Translations for the display and for the eventlog will follow

image

PR #1860, opened 2024-03-24
WIP: implement and use Utils::checkJsonOverflow()

schlimmchen:
this method calls the overflowed() method on the respective DynamicJsonDocument and prints a respective message if not all data could be added to the DynamicJsonDocument.

@tbnobody This is a proposal. I only used the new method in one spot. There are many more. Let me know what you think about this. I put this into OpenDTU-OnBattery (and used it on code specific to the downstream project) because I often find myself with problems in the web application that are caused by missing nodes in the respective JSON. I think is a valuable check to do right before serializing the JSON into a String buffer. Possibly, this new method can be changed to do the serialization and perform the check at the same time, so it becomes the standard to do both.

### DIFF
--- a/include/Utils.h
+++ b/include/Utils.h
@@ -11,5 +11,6 @@ public:
     static int getTimezoneOffset();
     static void restartDtu();
     static bool checkJsonAlloc(const DynamicJsonDocument& doc, const char* function, const uint16_t line);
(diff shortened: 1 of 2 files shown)


tbnobody:
I am currently prepare the migration to ArduinoJson 7. In this step I also re-implement the error handling (also for `AsyncJsonResponse`)

tbnobody:
Closing this PR as due to the update to ArduinoJson 7 all allocations are checked.

Issue #360, opened 2022-11-17
[Request] Web UI: better alignment for multiple inverters


scorpi11:
### Describe the solution you'd like

The alignment with multiple inverters looks somewhat strange if the names have different length.
<img width="465" alt="Bildschirmfoto 2022-11-17 um 12 23 41" src="https://user-images.githubusercontent.com/11080663/202434598-9b4b19cd-da98-4f0f-a2f9-a987c048fffb.png">

### Describe alternatives you've considered

The invterters should be evenly aligned independent from the length of the name string.

4c2:
I checked several of the possible other classes and possibilities. This is the best I could find: 
![grafik](https://github.com/tbnobody/OpenDTU/assets/76594587/71752686-b535-4a4f-817a-0d7a4d3bc677)

Most display sizes are fine, e.g.: 
![grafik](https://github.com/tbnobody/OpenDTU/assets/76594587/e79aba94-d11e-462c-a33c-7f26b220c5dc)

Only with very small size, I was not able to align it to the left:  
![grafik](https://github.com/tbnobody/OpenDTU/assets/76594587/1741ef3f-1984-47c6-9180-e47d2b45e078)

Any objections to implement this way? 
(if not, I would soon create a PR.)

PR #1870, opened 2024-03-25
webapp: better alignment inverter, issue 360

4c2:
fixing issue https://github.com/tbnobody/OpenDTU/issues/360

This is, how inverter view looks until now with some test strings: 
grafik

I checked several of the possible other classes and possibilities. This is way I like it most and which is realized in this PR: 
grafik

Most display sizes are fine, e.g.: 
grafik

Only with very small size, I was not able to align it to the left:  
grafik

Hint: I am not sure if I need to commit webapp_dist/js/app.js.gz as well or if it will be done by build process automatically. To be sure, I commited it as well.


tbnobody:
Looks good!
But could you please remove the app.js.gz from the commit and force push again?

4c2:
@tbnobody done.

tbnobody:
Had to squash it.... it's included in eff8d52014540b085eb84257fb4e178fe98b43af
Thank you!

PR #1909, opened 2024-04-12
remove remaining usage of F() macro

schlimmchen:
Given you removed most uses of said macro in d6028cbd, I guess these remaining four occurrences slipped through.


tbnobody:
You are right... I will include this PR in the next release.

PR #1920, opened 2024-04-18
Prevent compiling the whole project on each commit.

StefanOberhumer:
Putting the git information into a generated sourcefile prevents recompiling the whole project after each commit.


tbnobody:
Thank you. 

Would it make sense to put the `__compiled_constants.c` file into the `.gitignore` file as it will be regenerated on each compilation anyways?

StefanOberhumer:
`__compiled_constants.c` is generated in .pio/build/.... directory which is already excluded from git (included in `.gitignore`)

see:

https://github.com/tbnobody/OpenDTU/blob/97800434c4a7ad42fbca3915292ff3ab841a10c5/pio-scripts/auto_firmware_version.py#L59

tbnobody:
Ahhh sorry, yes you are right. My first assumption was, it gets generated in the `src` directory.

PR #2015, opened 2024-05-27
Feature: show ESP32 flash memory size in system info

schlimmchen:
This change adds a line to the system information of the web UI that lists the available amount of flash memory.

This info is vital for users of OpenDTU-OnBattery, as we have to insist on the availability of 8 MB of flash memory in the near future. This new line helps determine how much flash is actually available on a user's board.

Screenshot from 2024-05-27 21-55-42

Maybe this is interesting here as well?


tbnobody:
Yes! Thank you! I will include it in the next release

PR #2022, opened 2024-05-30
NFC: Includes list: Remove unneeded PinMapping.h

StefanOberhumer:
Also: Sort include directives by name.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 /*
- * Copyright (C) 2022 - 2023 Thomas Basler and others
+ * Copyright (C) 2022 - 2024 Thomas Basler and others
  */
+
 #include "Utils.h"
 #include "Display_Graphic.h"
 #include "Led_Single.h"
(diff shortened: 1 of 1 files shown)


tbnobody:
The `Pinmapping.h` can be removed, thats true. But the ordering is correct. The first include in the cpp file is always the corresponding header file (with the same name of the cpp file).

StefanOberhumer:
> ... But the ordering is correct. The first ....

Ah - OK!
I usually include all of the "standards" (#include **<**XX**>**) first!
But thinking about ... Your approach makes more sense ! :+1: 

Will adapt ...

tbnobody:
I've set the auto formatter style to `WebKit`. It sorts the corresponding header file at the first position followed by project specific headers and at least the global headers.

StefanOberhumer:
The now diff says:
I just removed the unneeded '#include PinMapping.h'  and changed a comment!

Based on your explanation: That result should match the project requirements.

Btrw: THX having comprehensible/explicable project settings/requirements :+1:

tbnobody:
Thank you and sorry for the nagging. I will include the PR in the next release.

PR #2025, opened 2024-06-02
check FW bin file size when creating factory.bin

schlimmchen:
In release 2024.05.03 of the downstream project OpenDTU-OnBattery, we accidentally released a firmware that would not even start on an ESP32 (non-S3). That was because the final `firmware.bin` was too large for the respective partition.

Since platformio spits out details on the flash usage when building, I always thought that the firmware is fine (regarding the size) as long as the build is successful. That was proven to be a wrong assumption by release 2024.05.03.

I did some research, and this does not seem to be a real headache among platformio users, probably because it is not very easy to create a firmware file that is too large for the partition but too little to trigger the platformio check?!

The reason is that platformio merely estimates the flash usage. The final firmware binary is slightly larger. Maybe that's because of additional sections of the ELF are being included that platformio did not consider when estimating the flash usage. Or maybe there is a header in the `firmware.bin` that platformio does not consider.

It's weird that platformio does not catch this even though it knows damn well the final partition size.

For mitigation, I added a check in the Python script which creates the `...factory.bin` file. I don't think this is ideal, this should be catched by platformio. I am also not confident that this is the best way to perform the check. Nevertheless, it's what I came up with. It trips for the 2024.05.03 tag of OpenDTU-OnBattery when building for `generic_esp32`, but not `generic_esp32s3`, as expected. It also does not trip if the firmware actually fits.

Maybe this is interesting here as well, even though this project has quite some marging left until hitting the flash partition limit.

### DIFF
--- a/pio-scripts/create_factory_bin.py
+++ b/pio-scripts/create_factory_bin.py
@@ -21,7 +21,7 @@ Import("env")
 platform = env.PioPlatform()
 
 import sys
-from os.path import join
+from os.path import join, getsize
 
 sys.path.append(join(platform.get_package_dir("tool-esptoolpy")))
 import esptool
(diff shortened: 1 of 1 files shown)


tbnobody:

PR #2052, opened 2024-06-10
avoid using pkg_resources

mpw96:
This is deprecated in python 3.12.
Also, improve file handling.


mpw96:
@tbnobody How do I notify you about this PR? Like this? ;-)

tbnobody:
Thank you! Will add it in the next release

PR #2077, opened 2024-06-19
Use long int format specifiers

hd1ex:
Long integers should be prefixed with l in format strings.

My compiler won't let me compile without fixing this.

### DIFF
--- a/include/defaults.h
+++ b/include/defaults.h
@@ -3,7 +3,7 @@
 
 #define SERIAL_BAUDRATE 115200
 
-#define APP_HOSTNAME "OpenDTU-%06X"
+#define APP_HOSTNAME "OpenDTU-%06lX"
 
 #define HTTP_PORT 80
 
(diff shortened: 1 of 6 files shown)


tbnobody:
If I apply your PR I get the following warnings:

```
src/NetworkSettings.cpp: In member function 'void NetworkSettingsClass::loop()':
src/NetworkSettings.cpp:210:38: warning: format '%ld' expects argument of type 'long int', but argument 3 has type 'uint32_t' {aka 'unsigned int'} [-Wformat=]
                 MessageOutput.printf("Admin AP remaining seconds: %ld / %ld\r\n", _adminTimeoutCounter, _adminTimeoutCounterMax);
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~
src/NetworkSettings.cpp:210:38: warning: format '%ld' expects argument of type 'long int', but argument 4 has type 'uint32_t' {aka 'unsigned int'} [-Wformat=]
```

What toolchain version are you using?

hd1ex:
Ok, seems like these types only change with the newer Arduino-ESP version. I will include these in my [change for supporting arduino-esp-v3](https://github.com/hd1ex/OpenDTU/commits/arduino-esp32-v3) and reorder them. Sorry for the noise!

PR #2115, opened 2024-06-30
use proper file handling

mpw96:
Just improving file handling:
- Close the file to read from immediately after reading,
- set encoding in open methods.

This actually fixes nothing, but makes the code [cleaner](https://refactoring.guru/refactoring/what-is-refactoring).

### DIFF
--- a/pio-scripts/patch_apply.py
+++ b/pio-scripts/patch_apply.py
@@ -23,10 +23,8 @@ def is_tool(name):
     return which(name) is not None
 
 def replaceInFile(in_file, out_file, text, subs, flags=0):
-    """
-        Function for replacing content for the given file
-        Taken from https://www.studytonight.com/python-howtos/search-and-replace-a-text-in-a-file-in-python
-    """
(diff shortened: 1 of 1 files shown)


tbnobody:
Please provide a proper description of this PR. It currently works as expected. Why do you want to change things?
Please also correct spelling mistakes.

mpw96:
> Please also correct spelling mistakes.
I wanted to write "massage", because I'm massaging the code... ;-)

mpw96:
@tbnobody If you don't like this change (and you are right that it fixes nothing) you can as well close this pull request. Thank you!

mpw96:
@tbnobody I have updated the pull request title and the description to explain why I think this change makes sense (basically: yes, the existing code works, but it's not as short and as understandable as possible).

mpw96:
@tbnobody This is just better python coding. It let's the library do the work. If you don't want it, please close the pull request, I'd of course be fine with it. Thank you!

PR #2117, opened 2024-06-30
#618 fix: explicitly disconnect prior connecting to wifi

jstammi:
See https://github.com/tbnobody/OpenDTU/issues/618#issuecomment-1661855347

It is a very old change in my installation, that I just noticed to have missed to open a pr for. Hopefully I remember the facts correctly.

The problem arises with the ap, being connected to, (temporarily) disappearing *) and requiring usage of a new "association packet" content (here: due to ap having changed the frequency).
The client wifi keeps the old one used initially (even across a soft-reset) and fails to connect to that ap then - endlessly.
To get it connected again actually power needs to be switched off. Or making the client "forgetting" this spcial packet prior trying to (re-)connect.

*) here, may be of interest, too, as I guess widely unknown: https://avm.de/service/wissensdatenbank/dok/FRITZ-Box-7490/1349_5-GHz-Funknetz-gelegentlich-deaktiviert-Radar-Signal/
Can be seen in the fritzbox events. To get around this, do not use automated frequency setting but fixed one outside radar ones.

### DIFF
--- a/src/NetworkSettings.cpp
+++ b/src/NetworkSettings.cpp
@@ -27,6 +27,8 @@ void NetworkSettingsClass::init(Scheduler& scheduler)
     WiFi.setScanMethod(WIFI_ALL_CHANNEL_SCAN);
     WiFi.setSortMethod(WIFI_CONNECT_AP_BY_SIGNAL);
 
+    WiFi.disconnect(true, true);
+
     WiFi.onEvent(std::bind(&NetworkSettingsClass::NetworkEvent, this, _1));
     setupMode();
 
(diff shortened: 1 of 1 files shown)


tbnobody:
I don't know how a behavior of the 5GHz WiFi band has impact of the ESP32 which only supports the 2.4GHz band. If I understand the AVM article correctly it only affects the 5GHz band.

jstammi:
You are correct, this problem was not on the SP32 but a different client, sry.

The problem wrt the "association packet" remains the same, just the reason for the ap disappearing is a different (e.g. firmware update). Unless this keeeping of the "association packet" was somewhere fixed in the underlaying libs, the fix should still be necessary (and made me entering the attic since having patched much less frequent).

jstammi:
The testing procedure is somehow easy:
- have the openDTU connected to the ap
- shutdown ap and start it again
- I expect the openDTU not being able to connect, I remember having seen those "association packet" messages in the serial log then
- I expect a soft-reset of the openDTU to not fix this, the connection should not be established to the initial ap again
- dis- and re-connect power to the openDTU, connection should be established again

PR #2168, opened 2024-07-23
Fix: inverter power limits precision

schlimmchen:
Hoymiles inverters allow setting relative limits with a precision of 0.1 %.
this changeset allows to utilize this precision.

* preserve accuracy when decoding power limit
* Web API: process floating point limits
* MQTT: process floating point limits
* use appropriate accuracy for limits in web UI
* HASS: step for relative inverter limit is 0.1 %


tbnobody:
... somehw merged with 70dacb5ea68630f0c59e244f83439846737cc027

PR #2198, opened 2024-08-08
fix: set sunset availability to false when its not available

AndreasBoehm:
While reading through the class to understand how we can use it to detect day and night in other parts of the project i stumbled across this line that is setting `_isSunsetAvailable` to `true` when we could not get the local time.

I assumed that this is a typo because I understood it that way that we can use `isSunsetAvailable()` to get if we can rely on `isDayPeriod()` or not.

Please let me know if my assumption is correct or if i misunderstood how this class works.

### DIFF
--- a/src/SunPosition.cpp
+++ b/src/SunPosition.cpp
@@ -73,7 +73,7 @@ void SunPositionClass::updateSunData()
     if (!gotLocalTime) {
         _sunriseMinutes = 0;
         _sunsetMinutes = 0;
-        _isSunsetAvailable = true;
+        _isSunsetAvailable = false;
         _isValidInfo = false;
         return;
     }


AndreasBoehm:
@tbnobody could you please take a look at this fix? Thanks

tbnobody:
Sunset Available means that a real sunset is available (e.g. in germany). In comparison to e.g. norway where a sunset / sunrise don't happen in summer month. This flag has nothing to do whether a value can be read. This flag mainly controls wether the sunset / sunrise time is shown in the web frontend. because if no sunset / sunrise is available (norway) the returned values do not say anything meaningfull. If this flag is false, the text `ntpinfo.NotAvailable` will be shown.

At this location in the source, the flag is set to true because we maybe want to see the current sunset/sunrise data even if no valid time is there. (Because if no sunrise/sunset is available we assume its day).

AndreasBoehm:
Thank your for the details! I will close this PR then.

PR #2293, opened 2024-09-21
Publish ESP heap and temperature details on MQTT

ranma:
I noticed that some useful ESP stats are missing on the MQTT feed, so this adds:

- ESP temperature
- ESP heap stats (size, free, minFree, maxAlloc)

image

### DIFF
--- a/src/MqttHandleDtu.cpp
+++ b/src/MqttHandleDtu.cpp
@@ -7,6 +7,7 @@
 #include "MqttSettings.h"
 #include "NetworkSettings.h"
 #include <Hoymiles.h>
+#include <CpuTemperature.h>
 
 MqttHandleDtuClass MqttHandleDtu;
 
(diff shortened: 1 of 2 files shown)


tbnobody:
Not sure why this happens, but the Home assistant values just show Unknown
image

The values are shown in the broker corretly:
image

tbnobody:
ahhh I see.... 
image


Wrong character at the end of the topic... will fix it

ranma:
> Wrong character at the end of the topic... will fix it

Oops, sorry. Thanks for the fix! I don't have HA set up for testing changes yet, I should do that... Also the rename from `minFree`/`maxAlloc` to `min_free`/`max_alloc` was only very lightly tested...

tbnobody:
I've already merged and corrected it. fdcbf9de9508aa769231595fd834527bc5e533c5

ranma:
> I've already merged and corrected it. [fdcbf9d](https://github.com/tbnobody/OpenDTU/commit/fdcbf9de9508aa769231595fd834527bc5e533c5)

Thanks!

PR #2331, opened 2024-10-05
Fix eslint issues

CommanderRedYT:
Fix for current CI problems


tbnobody:
Thank you for the PR. Could you please just prefix the commit message with "webapp: " to show it in the right category in the changelogs? Thank You!

CommanderRedYT:
@tbnobody Thanks, updated the commit name!

PR #2333, opened 2024-10-06
Remove EMAC related code for devices that don't have one

LennartF22:
It does not make sense to be able to configure Ethernet via the internal EMAC for targets other than the "original" ESP32, as they don't have one! Devices from the ESP32-C or ESP32-S series only support Ethernet via an external MAC, like the W5500. Therefore, the user should not be able to configure it and neither should see it in the pin configuration on the "Device-Manager" page of the OpenDTU web interface.

Furthermore, the usage of `ETH_PHY_LAN8720` in the `PinMappingClass` caused a compilation error when using Arduino core v3.0, because with v3.0 the definition only exists for targets where internal EMAC support is configured (see [`ETH.h`](https://github.com/espressif/arduino-esp32/blob/0a554933d0304b4d6a5a05f5762f5ce7368cc70e/libraries/Ethernet/src/ETH.h#L101)).

### DIFF
--- a/include/PinMapping.h
+++ b/include/PinMapping.h
@@ -12,6 +12,7 @@
 
 struct PinMapping_t {
     char name[MAPPING_NAME_STRLEN + 1];
+
     int8_t nrf24_miso;
     int8_t nrf24_mosi;
     int8_t nrf24_clk;
@@ -33,6 +34,7 @@ struct PinMapping_t {
     int8_t w5500_int;
     int8_t w5500_rst;
 
+#if CONFIG_ETH_USE_ESP32_EMAC
     int8_t eth_phy_addr;
     bool eth_enabled;
     int eth_power;
@@ -40,11 +42,14 @@ struct PinMapping_t {
     int eth_mdio;
     eth_phy_type_t eth_type;
     eth_clock_mode_t eth_clk_mode;
+#endif
+
     uint8_t display_type;
     uint8_t display_data;
     uint8_t display_clk;
     uint8_t display_cs;
     uint8_t display_reset;
+
     int8_t led[PINMAPPING_LED_COUNT];
 };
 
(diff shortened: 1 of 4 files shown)


schlimmchen:
@LennartF22 @tbnobody FYI: https://github.com/helgeerbe/OpenDTU-OnBattery/commit/d6d5f32329731ffd16585c200bf6d424069a94c7

tbnobody:
Haven't tested it yet, but what happens if pins are defined via define and the pin mapping contains a empty or non-exsiting eth section?

schlimmchen:
You will then (still) see the category "eth" in the web UI if and only if the firmware was compiled with `CONFIG_ETH_USE_ESP32_EMAC`, as also the `#define`s for the respective pins are guarded by this `#define`.

Or in general: A category will appear in the device manager web UI if it is included in `curPin` at the `api/device/config` endpoint, regardless of whether or not it is in the pin mapping. Sections in the pin mapping are masked if the firmware says "I don't know this category of pins", which I think makes sense in general.

LennartF22:
> Haven't tested it yet, but what happens if pins are defined via define and the pin mapping contains a empty or non-exsiting eth section?

The defines from the `platformio.ini` are ignored, since they are never used due to this patch. However, when Ethernet pins are defined in the `pin_mapping.json`, they will be shown in the web interface, because - as far as I know - the web interface will happily show anything configured in the `pin_mapping.json`.

schlimmchen:
> because - as far as I know - the web interface will happily show anything configured in the pin_mapping.json.

That's exactly what the one-liner in https://github.com/helgeerbe/OpenDTU-OnBattery/commit/d6d5f32329731ffd16585c200bf6d424069a94c7 prevents. The categories (and only the categories, not the individual pins) are filtered by the known/supported categories of the firmware.

LennartF22:
> That's exactly what the one-liner in [helgeerbe@d6d5f32](https://github.com/helgeerbe/OpenDTU-OnBattery/commit/d6d5f32329731ffd16585c200bf6d424069a94c7) prevents. The categories (and only the categories, not the individual pins) are filtered by the known/supported categories of the firmware.

Yes, sorry, I only noticed that @tbnobody‘s question was directed towards your change after writing my comment, but I already updated it ^^

PR #2360, opened 2024-10-17
PoC: config write guard

schlimmchen:
I was trying to tackle application of changed Dynamic Power Limiter settings in my feature branch implementing support for multiple inverters managed by the DPL. Then I noticed, that a bunch of nasty problem would go away if I knew that I could trust the config being consistent, i.e., if the config was only written synchronously with the main loop.

The latter ist currently not true: The config is usually manipulated in the context of the async webserver task, while it is read in the context of the main loop.

I wanted to keep the overhead when reading the config to a minimum, so locking a mutex each time the config shall be read is out of the picture, IMHO.

Writing the config happens rather seldom. So it should be okay if that takes a couple of milliseconds longer -- in the worst case.

My approach: Writing the config is only possible when acquiring a WriteGuard, which locks the "config write mutex" and waits for the signal (from the main loop) to start writing. The signal is needed to make writing the config synchronous with the main loop, and the mutex is needed to use the condition_variable but also to lock multiple writers from each other (which is less of a concern, to be honest).

I like this approach, because I like RAII. However, I am not sure if this is the best approach. Another idea would be to enqueue a callback which is then executed within the context of the main loop. We would need to copy and bind to that callback a significant amount of data, which seems off-putting. Also, access to the queue of callbacks then also need to be locked by a mutex.

@tbnobody I would like to know your opinion on this. Do you like this approach?

@LennartF22 We talked about this briefly. What are your thoughts on this approach?

### DIFF
--- a/include/Configuration.h
+++ b/include/Configuration.h
@@ -3,6 +3,9 @@
 
 #include "PinMapping.h"
 #include <cstdint>
+#include <TaskSchedulerDeclarations.h>
+#include <mutex>
+#include <condition_variable>
 
 #define CONFIG_FILENAME "/config.json"
 #define CONFIG_VERSION 0x00011c00 // 0.1.28 // make sure to clean all after change
@@ -161,15 +164,32 @@ struct CONFIG_T {
 
 class ConfigurationClass {
 public:
-    void init();
+    void init(Scheduler& scheduler);
     bool read();
     bool write();
     void migrate();
-    CONFIG_T& get();
+    CONFIG_T const& get();
+
+    class WriteGuard {
+    public:
+        WriteGuard();
+        CONFIG_T& getConfig();
+        ~WriteGuard();
+
+    private:
+        std::unique_lock<std::mutex> _lock;
+    };
+
+    WriteGuard getWriteGuard();
 
     INVERTER_CONFIG_T* getFreeInverterSlot();
     INVERTER_CONFIG_T* getInverterConfig(const uint64_t serial);
     void deleteInverterById(const uint8_t id);
+
+private:
+    void loop();
+
+    Task _loopTask;
 };
 
 extern ConfigurationClass Configuration;
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -13,8 +13,17 @@
 
 CONFIG_T config;
 
-void ConfigurationClass::init()
+static std::condition_variable sWriterCv;
+static std::mutex sWriterMutex;
+static unsigned sWriterCount = 0;
+
+void ConfigurationClass::init(Scheduler& scheduler)
 {
+    scheduler.addTask(_loopTask);
+    _loopTask.setCallback(std::bind(&ConfigurationClass::loop, this));
+    _loopTask.setIterations(TASK_FOREVER);
+    _loopTask.enable();
+
     memset(&config, 0x0, sizeof(config));
 }
 
@@ -318,6 +327,20 @@ bool ConfigurationClass::read()
     }
 
     f.close();
+
+    // Check for default DTU serial
+    MessageOutput.print("Check for default DTU serial... ");
+    if (config.Dtu.Serial == DTU_SERIAL) {
+        MessageOutput.print("generate serial based on ESP chip id: ");
+        const uint64_t dtuId = Utils::generateDtuSerial();
+        MessageOutput.printf("%0" PRIx32 "%08" PRIx32 "... ",
+            ((uint32_t)((dtuId >> 32) & 0xFFFFFFFF)),
+            ((uint32_t)(dtuId & 0xFFFFFFFF)));
+        config.Dtu.Serial = dtuId;
+        write();
+    }
+    MessageOutput.println("done");
+
     return true;
 }
 
@@ -390,9 +413,14 @@ void ConfigurationClass::migrate()
     read();
 }
 
-CONFIG_T& ConfigurationClass::get()
+CONFIG_T const& ConfigurationClass::get()
 {
     return config;
+}
+
+ConfigurationClass::WriteGuard ConfigurationClass::getWriteGuard()
+{
+    return WriteGuard();
 }
 
 INVERTER_CONFIG_T* ConfigurationClass::getFreeInverterSlot()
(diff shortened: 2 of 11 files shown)


LennartF22:
@schlimmchen I have not looked at the changes in detail yet, but it's definitely an improvement.

There are more synchronization issue between the WebApi and the main loop though, like here (if I'm not mistaken): https://github.com/tbnobody/OpenDTU/blob/dc5eb96f5035ba9e058e38d83df2d6c691b0764b/src/WebApi_inverter.cpp#L144

tbnobody:
Looks great! I am just curious if we would ever have to write the config in the context of the main loop (as it will generate a dead lock). Currently I cannot imagine any use case.

schlimmchen:
Thanks for your feedback!

> I am just curious if we would ever have to write the config in the context of the main loop (as it will generate a dead lock).

Yes. I saw that, too. If one would inadvertently try to do this, the deadlock would happend when testing the change for the first time, and the config would not be saved at all, so it should be noticed early. And if we need to do that, we can adjust the WriteGuard c'tor to see whether the current task it the main loop task and then only lock the mutex, not waiting for the signal.

Let me polish and double-check this, then I will un-draft this PR (next week).

schlimmchen:
@tbnobody are you passionate about

`auto const& config = Configuration.get();`

versus 

`const auto& config = Configuration.get();`

or whether it's consistent across the code base?

schlimmchen:
I am still happy with this. I double checked that generating a DTU serial works as expected using an erased ESP32:

```
[2024-10-22 21:27:07.684] rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
[2024-10-22 21:27:07.690] configsip: 0, SPIWP:0xee
[2024-10-22 21:27:07.690] clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[2024-10-22 21:27:07.695] mode:DIO, clock div:2
[2024-10-22 21:27:07.701] load:0x3fff0030,len:1184
[2024-10-22 21:27:07.701] load:0x40078000,len:13232
[2024-10-22 21:27:07.707] load:0x40080400,len:3028
[2024-10-22 21:27:07.707] entry 0x400805e4
[2024-10-22 21:27:08.379] E (694) esp_core_dump_flash: No core dum���ѥѥ���found!
[2024-10-22 21:27:08.379] E (694) esp_core_dump_flash: No core dump partition found!
[2024-10-22 21:27:08.398] 
[2024-10-22 21:27:08.398] Starting OpenDTU
[2024-10-22 21:27:08.398] Initialize FS... E (13) esp_littlefs: ./components/esp_littlefs/src/littlefs/lfs.c:1367:error: Corrupted dir pair at {0x0, 0x1}
[2024-10-22 21:27:08.409] 
[2024-10-22 21:27:08.409] E (20) esp_littlefs: mount failed,  (-84)
[2024-10-22 21:27:08.415] E (23) esp_littlefs: Failed to initialize LittleFS
[2024-10-22 21:27:08.415] [    37][E][LittleFS.cpp:98] begin(): Mounting LittleFS failed! Error: -1
[2024-10-22 21:27:08.426] failed... trying to format...E (46) esp_littlefs: ./components/esp_littlefs/src/littlefs/lfs.c:1367:error: Corrupted dir pair at {0x0, 0x1}
[2024-10-22 21:27:08.437] 
[2024-10-22 21:27:08.437] E (47) esp_littlefs: mount failed,  (-84)
[2024-10-22 21:27:08.437] E (51) esp_littlefs: Failed to initialize LittleFS
[2024-10-22 21:27:08.475] failedReading configuration... [   105][E][vfs_api.cpp:105] open(): /littlefs/config.json does not exist, no permits for creation
[2024-10-22 21:27:08.486] Failed to read file, using default configuration
[2024-10-22 21:27:08.486] Check for default DTU serial... generate serial based on ESP chip id: 199980103760... done
[2024-10-22 21:27:08.754] done
[2024-10-22 21:27:08.754] Reading PinMapping... [   384][E][vfs_api.cpp:105] open(): /littlefs/pin_mapping.json does not exist, no permits for creation
[2024-10-22 21:27:08.767] using default config done
[2024-10-22 21:27:08.767] Initialize Network... done
[2024-10-22 21:27:09.039] Setting Hostname... Initialize NTP... done
[2024-10-22 21:27:09.039] Initialize SunPosition... done
[2024-10-22 21:27:09.044] Initialize MqTT... done
[2024-10-22 21:27:09.044] Initialize WebApi... done
[2024-10-22 21:27:09.054] Initialize Display... done
[2024-10-22 21:27:09.054] Initialize LEDs... done
[2024-10-22 21:27:09.054] Initialize Hoymiles interface... Invalid pin config
[2024-10-22 21:27:09.060] Switch to WiFi mode
[2024-10-22 21:27:09.064] Setting Hostname... done
[2024-10-22 21:27:39.403] Disable search for AP... done
```

include/Configuration.h
```diff
@@ -161,15 +164,32 @@ struct CONFIG_T {
 
 class ConfigurationClass {
 public:
-    void init();
+    void init(Scheduler& scheduler);
     bool read();
     bool write();
     void migrate();
-    CONFIG_T& get();
+    CONFIG_T const& get();
```
mathieucarbou, line 171: this getter is problematic: how it ensures that reads going through it will be protected by the mutex ?
What if another task / core modifies the config while another one calls this method to access a value ?

schlimmchen: yes, there is no assurance. I did not want to create overhead when reading the config structure. we assume the config structure is only read in context of the main loop. since writing is synchronized with the main loop using the changes in this PR, there is no issue when reading the structure in context of the main loop. however, other tasks also do read the structure, and indeed, those reads are still not protected through the changes in this PR. IMHO, that's a next step. it's going to be a pain to address this...

mathieucarbou: No that's not complicated: the config class just has to expose the mutex (or 2 methods) and where you read the config in the code, you lock the mutex before then unlock it.

The ESP-IDF and Arduino Core are full of designs like that, especially in the LWIP layer which now require to lock/unlock all operations since Matter was introduced.

schlimmchen: I did not write it was complicated, just a pain :wink: What I meant was: large changeset. If we want to use a lock guard as well, we also need an additional scope, meaning indenting code, meaning even larger changeset. And things do indeed get complicated, or at least nasty, if the mutex must be released before a function is called, as to avoid recursive locking. I assume recursive locks are to be avoided, especially since we also use other mutexes and deadlocks are looming once we need to juggle multiple locks simultaneously.

> The ESP-IDF and Arduino Core are full of designs like that

You are suggesting that we should not be afraid of the overhead of locking the same mutex dozens of times in the main loop in every round of the loop? Usually, I am very liberal when it comes to trading some performance for new and useful features, but locking mutexes? Am I paranoid?

mathieucarbou: I agree! 
All depends on who reads, when and at which frequency. 
I don't know enough the app internally.
Maybe otherwise returning copies or partial copies of the config when reading would be an option also 
I was just highlighting the fact that there is no protection theorically.
But if in practice it never happens then it's good.

PR #2387, opened 2024-10-29
Fix: skip BOM in JSON files (pin_mapping and config)

schlimmchen:
For obvious reasons (I guess).

See hoylabs#1355.


tbnobody:
I've implemented it as a function instead of duplicated code in 55c98ef880a900085cc6c0cb9cd59833af3720c9

schlimmchen:
Sure, makes sense! 👍

PR #2457, opened 2024-12-14
Merge with v24.11.7

Benichou34:


schlimmchen:
@Benichou34 You have my sympathies, I made this mistake at least twice :grin:

PR #2584, opened 2025-03-11
Fix Home Assistant Energy Management

functionpointer:
YieldTotal in Home Assistant Energy Management is somewhat broken, due to YieldTotal intermittently jumping to 0 or incomplete values from a subset of inverters. This happens after reboots, especially at night, and also sometimes during connectivity issues.

This PR fixes it by introducing availability topics to most MQTT topics.
Each inverter's topics are only available if that inverter is reachable.

- The total topics `ac/yieldtotal` and `ac/yieldday` are only available when *all* inverters are reachable.
- Bugs in `isReachable()` and `isAllEnabledReachable()` are fixed.
- When connecting to MQTT, the total topics are forcefully set to `unavailable` for a few seconds. This ensures no invalid data is accidentally taken as valid. Previously, rebooting would cause `ac/yieldtotal`=0 to be published *before* `ac/is_valid` = false.

### DIFF
--- a/include/MqttHandleInverterTotal.h
+++ b/include/MqttHandleInverterTotal.h
@@ -2,6 +2,8 @@
 #pragma once
 
 #include <TaskSchedulerDeclarations.h>
+#include "MqttAvailableHandler.h"
+#include <memory>
 
 class MqttHandleInverterTotalClass {
 public:
@@ -10,10 +12,11 @@ public:
 
 private:
     void loop();
+    bool isDataValid();
+    void sendData();
 
     Task _loopTask;
-
-    unsigned int _connected_iterations = 0;
+    std::unique_ptr<MqttAvailableHandler> _availableHandler;
 };
 
 extern MqttHandleInverterTotalClass MqttHandleInverterTotal;
--- a/include/MqttHandleInverterTotal.h
+++ b/include/MqttHandleInverterTotal.h
@@ -12,6 +12,8 @@ private:
     void loop();
 
     Task _loopTask;
+
+    unsigned int _connected_iterations = 0;
 };
 
 extern MqttHandleInverterTotalClass MqttHandleInverterTotal;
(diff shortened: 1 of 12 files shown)


src/Datastore.cpp
```diff
@@ -76,7 +76,7 @@ void DatastoreClass::loop()
         if (inv->isReachable()) {
             isReachable++;
         } else {
-            if (inv->getEnablePolling()) {
+            if (cfg->Poll_Enable) {
```
schlimmchen, line 79: Why is this necessary here but not elsewhere above and below?

include/MqttHandleInverterTotal.h
```diff
@@ -12,6 +12,8 @@ class MqttHandleInverterTotalClass {
     void loop();
 
     Task _loopTask;
+
+    unsigned int _connected_iterations = 0;
```
schlimmchen, line 16: Style: This should be CamelCase.

lib/Hoymiles/src/inverters/InverterAbstract.cpp
```diff
@@ -74,7 +74,7 @@ bool InverterAbstract::isProducing()
 
 bool InverterAbstract::isReachable()
 {
-    return _enablePolling && Statistics()->getRxFailureCount() <= _reachableThreshold;
+    return _enablePolling && Statistics()->getRxFailureCount() <= _reachableThreshold && Statistics()->getLastUpdate() > 0;
```
schlimmchen, line 77: This makes sense and should have gone into its own PR :wink:

src/Datastore.cpp
tbnobody: The original logic was to set all available if they are suposed to be available. That means if you don't fetch inverter values at night, all other inverters are still valid because you intentional don't pull this specific inverter. 
Example: You have two inverters, one at a solar panel, one at a battery (which can run the whole night). If you don't fetch the first one at night, the ac_total should be still valid because the second inverter works as expected (as so does the first one as it is suspected to be offline)

delacor:
any news here?

PR #2633, opened 2025-04-08
bump ESPAsyncWebServer to v3.7.6

plieven:
+ dynamically adjust prometheus endpoint response buffer size

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -42,7 +42,7 @@ build_unflags =
     -std=gnu++11
 
 lib_deps =
-    ESP32Async/ESPAsyncWebServer @ 3.7.4
+    ESP32Async/ESPAsyncWebServer @ 3.7.6
     bblanchon/ArduinoJson @ 7.3.1
     https://github.com/bertmelis/espMqttClient.git#v1.7.0
     nrf24/RF24 @ 1.4.11
--- a/src/WebApi_prometheus.cpp
+++ b/src/WebApi_prometheus.cpp
@@ -20,12 +20,14 @@ void WebApiPrometheusClass::init(AsyncWebServer& server, Scheduler& scheduler)
 
 void WebApiPrometheusClass::onPrometheusMetricsGet(AsyncWebServerRequest* request)
 {
+    static size_t initialResponseBufferSize = 1024;
+
     if (!WebApi.checkCredentialsReadonly(request)) {
         return;
     }
 
     try {
-        auto stream = request->beginResponseStream("text/plain; charset=utf-8", 40960);
+        auto stream = request->beginResponseStream("text/plain; charset=utf-8", initialResponseBufferSize);
 
         stream->print("# HELP opendtu_build Build info\n");
         stream->print("# TYPE opendtu_build gauge\n");
@@ -109,6 +111,10 @@ void WebApiPrometheusClass::onPrometheusMetricsGet(AsyncWebServerRequest* reques
             }
         }
         stream->addHeader("Cache-Control", "no-cache");
+        if (stream->available() > initialResponseBufferSize) {
+            initialResponseBufferSize = stream->available();
+            MessageOutput.printf("Increased /api/prometheus/metrics initialResponseBufferSize to %" PRIu32 " bytes\r\n", initialResponseBufferSize);
+        }
         request->send(stream);
 
     } catch (std::bad_alloc& bad_alloc) {


plieven:
fixed, I should have known better ...

tbnobody:
I see a `Failed to allocate` on each request:
```
[374549][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
Increased /api/prometheus/metrics initialResponseBufferSize to 1863 bytes<\r><\n>
[384389][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
[385314][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
[386019][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
```

plieven:
@mathieucarbou  i think the check in ESPAsyncWebserver is wrong. I guess cbuf allocates an extra byte for a terminating \0.

That check should be >= and not ==.

tbnobody:
I think I found the reason...
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/freertos_additions.html
image

mathieucarbou:
crap!

mathieucarbou:
Why I didn't get the warning when I run my MRE ?

mathieucarbou:
@tbnobody : weird, I do not have teh same cbuf impl. as you 🤯  I do not have the extra character. 
You are using latest arduin core ?

mathieucarbou:
> @mathieucarbou i think the check in ESPAsyncWebserver is wrong. I guess cbuf allocates an extra byte for a terminating \0. That check should be >= and not ==.

More like `if (bufferSize && _content->size() < bufferSize) {`

plieven:
> > @mathieucarbou i think the check in ESPAsyncWebserver is wrong. I guess cbuf allocates an extra byte for a terminating \0. That check should be >= and not ==.
> 
> More like `if (bufferSize && _content->size() < bufferSize) {`

Yes, I replied via e-mail and haven't checked the source... The change you propose should work with both core versions. Just to be 100% sure if the initial allocation fails, cbuf isn't NULL or is it?

plieven:
> I see a `Failed to allocate` on each request:
> 
> ```
> [374549][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
> Increased /api/prometheus/metrics initialResponseBufferSize to 1863 bytes<\r><\n>
> [384389][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
> [385314][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
> [386019][E][WebResponses.cpp:827] AsyncResponseStream(): Failed to allocate<\r><\n>
> ```
> 
> I am pretty sure that's new.

thats only visible on the serial console, right? I only checked the webconsole though.

mathieucarbou:
> Yes, I replied via e-mail and haven't checked the source... The change you propose should work with both core versions. Just to be 100% sure if the initial allocation fails, cbuf isn't NULL or is it?

cbuf cannot be null: it is not a `new (std::nothrow)`. If cbuf cannot be created, it will fail. 

On Arduino Core 3, not likely because the call to `xRingbufferCreate(size, RINGBUF_TYPE_BYTEBUF)` in the ctor will return a null ring buffer, so the whole class will act as a no-op.

 On Arduino Core 2, there is a new chat[] in the ctor, so the ctor will crahs the board on a failed allocation because of the usage of `new`. Nothing that we can not - this is arduino core code.

OpenDTU should switch to Arduino Core 3 with pioarduino: 

`platform = https://github.com/pioarduino/platform-espressif32/releases/download/54.03.20/platform-espressif32.zip`

mathieucarbou:
> thats only visible on the serial console, right? I only checked the webconsole though.

that's because of the arduino core 2 implementation not returning the same size as arduino core 3 implementation.

this is just an annoying error log: the cbuf is allocated behind.

plieven:
Okay, understood. In Core 2 it will crash, but in Core 3 the buffer can be up to 3 bytes larger as @tbnobody mentioned from the docs. So we need the modified check?

plieven:
okay, what are the official requirements for ESPAsyncWebServer? If you allow arduino core 2, you should fix that bogus error message.

While looking at WebResponses.cpp I checked what happens in AsyncResponseStream::write if the resize fails. It will return the orignal size. We do not detect this and just overwrite contents of the ring buffer if it wrappes. Do not know what the right behaviour is. return 0 and do nothing, return the requested write size and do nothring or overwrite...

mathieucarbou:
> ESPAsyncWebServer

We support latest released versions of Arduino Core 2 and 3, but we strongly advise to use 3 because at one point we won't be able to keep 2 compatibility.

Arduino Core team does not maintain 2 anymore: this is a EOL version.

Everybody should switch to 3 now....

mathieucarbou:
I have a PR ready that is building and will release a v3.7.6 in a few minutes: https://github.com/ESP32Async/ESPAsyncWebServer/pull/152

plieven:
> I have a PR ready that is building and will release a v3.7.6 in a few minutes: [ESP32Async/ESPAsyncWebServer#152](https://github.com/ESP32Async/ESPAsyncWebServer/pull/152)

Yeah, the commit creeped in the conversation so I missed it.

mathieucarbou:
> While looking at WebResponses.cpp I checked what happens in AsyncResponseStream::write if the resize fails. It will return the orignal size. We do not detect this and just overwrite contents of the ring buffer if it wrappes. Do not know what the right behaviour is. return 0 and do nothing, return the requested write size and do nothring or overwrite...

Oh !!! Yes that should be fixed... No overwrite should happen. Is it on Arduino core 2 ?

plieven:
@mathieucarbou v2 and v3 are in agreement on that:

https://github.com/espressif/arduino-esp32/blob/b333bf2697463f76e98640d4421dfd1560482c1f/cores/esp32/cbuf.cpp#L92

https://github.com/espressif/arduino-esp32/blob/1f038677eb2eaf5e9ca6b6074486803c15468bed/cores/esp32/cbuf.cpp#L52

plieven:
Fortunately both implementations clamp the write size to the buffer size:

`size_t size_to_write = (size < bytes_available) ? size : bytes_available;`

mathieucarbou:
Added a commit to log also the failed re-allocations: https://github.com/ESP32Async/ESPAsyncWebServer/pull/152/commits/bf9f457bb8066dab2527c2da468cd97760c02d74

plieven:
@tbnobody updated the PR to include ESPAsyncWebServer v3.7.6. The arduino core update to v3 is a different thing that should be thoughtfully tested.

mathieucarbou:
> AsyncJsonResponse

`AsyncJsonResponse` has its own implementation: it is not using AsyncStreamResponse. You can use AsyncStreamResponse to send a JsonDocument because AsyncStreamResponse is a Print, but AsyncJsonResponse has its own tricky implementation:

```
size_t PrettyAsyncJsonResponse::_fillBuffer(uint8_t *data, size_t len) {
  ChunkPrint dest(data, _sentLength, len);
#if ARDUINOJSON_VERSION_MAJOR == 5
  _root.prettyPrintTo(dest);
#else
  serializeJsonPretty(_root, dest);
#endif
  return len;
}
```

The ChunkPrint is a hack to skip the first bytes if _fillBuffer is called again  (which triggers a new json serialization).

What I do not understand though in this code is why len is returned instead of `return min(len, serializeJson(_root, dest));`

PR #2905, opened 2025-12-05
ESP32Async/ESPAsyncWebServer @ 3.10.3

mathieucarbou:
Upgrade PR for:
- ESP32Async/ESPAsyncWebServer @ 3.10.3
- ESP32Async/AsyncTCP @ 3.4.10

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -49,8 +49,8 @@ build_unflags =
     -std=gnu++11
 
 lib_deps =
-    ESP32Async/ESPAsyncWebServer @ 3.8.0
-    ESP32Async/AsyncTCP @ 3.4.7
+    ESP32Async/ESPAsyncWebServer @ 3.10.3
+    ESP32Async/AsyncTCP @ 3.4.10
     bblanchon/ArduinoJson @ 7.4.2
     https://github.com/bertmelis/espMqttClient.git#v1.7.0
     nrf24/RF24 @ 1.4.11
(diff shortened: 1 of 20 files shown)


tbnobody:
I have already merged this to my local dev branch and also updated to 3.9.3
In this branch I also included newest versions for the webapp. I will maybe release before x-mas.

And thank you for your efforts

mathieucarbou:
@tbnobody : FYI I have updated the PR with the new versions if you want to grab it locally also. 

3.10.0 substantially improved WebSocket communication to move as much data as possible within the client pcb space, which leads to less messages sitting on heap in the queue and insane perf improvements in terms of message delivery. This is a big improvements especially when sending short messages.

=> https://github.com/ESP32Async/ESPAsyncWebServer/releases

Also these 2 could be removed:

```
    -DCONFIG_ASYNC_TCP_EVENT_QUEUE_SIZE=128
    -DCONFIG_ASYNC_TCP_QUEUE_SIZE=128
```

=> `CONFIG_ASYNC_TCP_EVENT_QUEUE_SIZE` does not exist in AsyncTCP

=> `CONFIG_ASYNC_TCP_QUEUE_SIZE` (queue size for lwip events) default is 64. In therory, lwip queue size should be small (1 or 2 events per client max). If it grows, it means the application has an issue: it is blocking for too long the async_tcp task because it executes long-running code in a callback of async_tcp. In that case, this is a use case for an app refactoring, WebSocket usage, or use the new request continuation feature that allows to pause a request.

These parameters were probably more valid with the old AsyncTCP but it has underwent a lot of improvements lately.

tbnobody:
I think we can close this for now. 
Thank you for all your work and efforts which you put into this webserver.

PR #582, opened 2023-02-13
Added Modbus to OpenDTU

ArekKubacki:
Added Modbus TCP/IP

PR #2962, opened 2026-02-01
feat: Add MIT-5000-8T (MIT_8CH) inverter support + RF capture mode

### LINKED ISSUE #2742: [Request] Support of new Hoymiles Inverter-Series MIT-4000/4500/5000-8T

I'm currently using a Hoymiles HMT-2000 3-phase Inverter very successfully controlled by the "onBattery" fork of OpenDTU. With the intention to increase the Inverter power I thought about adding another HMT-2000 to my setup or to replace the HMT-2000 with a new MIT-4000-8T. Although the new inverter also uses sub 1G communication, it is not yet supported by OpenDTU. But I would prefer the single Inverter solution because I expect that it can be better controlled than two Inverters.

Please add the support for the new Hoymiles Inverter-Series MIT-4000/4500/5000-8T.

_No response_

### Additional context

If there is something I can do to support the implementation of the enhancement please let me know.
I could buy one inverter and could support communication recording (if I'm told how to do this).

Geoffn-Hub:
## Summary

This PR adds native support for the **Hoymiles MIT-5000-8T** 3-phase microinverter and includes an **RF capture mode** tool used to reverse-engineer the protocol.

Closes #2742

### Key Findings

- The MIT-5000-8T (serial prefix `0x1520`) reports **4 independent MPPT channels** over RF, despite the "8T" name — each MPPT tracker has 2 panel inputs sharing a single tracker
- The inverter follows the HMT 3-phase pattern for AC fields (line-neutral voltages, line-line voltages, per-phase currents, power factor)
- DC layout: 12 bytes per channel (UDC/IDC/PDC/YT[4]/YD)
- Total payload: 92 bytes across 6 RF frames (01-05 + 0x86), reassembling to 88 data + 2 CRC16

### Protocol Behaviour Differences from HMS/HMT

Testing with a live MIT-5000-8T revealed several protocol differences that required OpenDTU changes:

1. **SystemConfigPara response is 38 bytes** (vs 48 for HMS/HMT). The hardcoded `SYSTEM_CONFIG_PARA_SIZE` check was rejecting valid responses. Fix: made expected byte count configurable via `setExpectedByteCount()`, with MIT_8CH setting 38 in its constructor.

2. **The MIT sends only 2 fragments per response burst**, then waits for individual retransmit requests for each remaining fragment. The HMS/HMT sends all fragments in a single burst. Each retransmit cycle requires ~3 requests before the MIT responds. With 6 total fragments and 4 needing individual recovery at ~3 retransmits each, the default `MAX_RETRANSMIT_COUNT` of 5 was insufficient. Increased to 20 for reliable reception.

3. **Retransmit RX window**: The `SingleDataCommand` (RequestFrame) timeout was increased from 100ms to 250ms to accommodate the MIT's slower response timing.

**Result**: MIT RealTimeRunData now achieves ~100% success rate, with each poll cycle taking ~4-5 seconds due to the retransmit dance.

### Validated Byte Assignments

All byte assignments were validated by cross-referencing passive RF captures from the MIT-5000-8T against DTU-Pro Modbus TCP data read via Home Assistant sensors:

| Field | RF Value | HA Value | Match |
|-------|----------|----------|-------|
| MPPT1 YT | 80.368 kWh | ~80.37 kWh | ✅ |
| MPPT2 YT | 65.066 kWh | ~65.07 kWh | ✅ |
| MPPT3 YT | 155.147 kWh | ~155.15 kWh | ✅ |
| MPPT4 YT | 151.566 kWh | ~151.57 kWh | ✅ |
| Frequency | 50.00 Hz | 50.00 Hz | ✅ |
| PAC | 225.4 W | ~230 W | ✅ |
| Temperature | 12.0 °C | 12.0 °C | ✅ |
| Power Factor | 1.000 | 1.000 | ✅ |

### Complete Decoded Capture (230W production, 2026-02-01)

Raw frames:
```
0x01: 01 00 01 01 51 00 67 01 5D 00 01 39 F0 00 25 01 62
0x02: 02 00 66 01 70 00 00 FE 2A 00 23 02 D3 00 7E 03 91
0x03: 03 00 02 5E 0B 00 58 02 B5 00 6F 03 03 00 02 50 0E
0x04: 04 00 64 09 0D 09 1C 09 1F 0F AB 0F D2 0F C1 13 88
0x05: 05 00 00 08 CE 00 00 00 E1 00 22 00 21 00 1E 03 E8
0x86: 86 00 78 00 03 00 00 00 00 00 1A 63 40
```

### Unresolved Fields

- **Offset 68**: Always 0 in captures — likely reserved
- **Offset 70**: Value matches PAC exactly — possibly apparent power (S in VA). Needs capture under reactive load to confirm.

### Files Changed (MIT Support)

- `lib/Hoymiles/src/inverters/MIT_8CH.cpp` — Inverter implementation with validated byte assignments
- `lib/Hoymiles/src/inverters/MIT_8CH.h` — Header
- `lib/Hoymiles/src/Hoymiles.cpp` — Register MIT_8CH for serial prefix `0x1520`
- `lib/Hoymiles/src/parser/SystemConfigParaParser.cpp` — Configurable expected byte count
- `lib/Hoymiles/src/parser/SystemConfigParaParser.h` — `setExpectedByteCount()` method
- `lib/Hoymiles/src/commands/CommandAbstract.h` — `MAX_RETRANSMIT_COUNT` increased to 20
- `lib/Hoymiles/src/commands/SingleDataCommand.cpp` — RequestFrame timeout 100ms → 250ms

## RF Capture Mode

A diagnostic tool for reverse-engineering new inverter protocols by passively sniffing RF traffic.

### How It Works

1. **Channel hopping**: CMT2300A radio sweeps all legal channels (863-870 MHz EU) with 50ms dwell time
2. **Frame logging**: All valid CRC8 frames are logged to the serial console (WebSocket) with source/destination serial, RSSI, frequency, and hex dump
3. **Non-destructive**: Operates alongside normal inverter polling infrastructure

### API

```bash
# Enable
curl -X POST "http://admin:<pass>@<dtu-ip>/api/dtu/capture" \
  --data-urlencode 'data={"capture_mode":true}'

# Status
curl -s "http://admin:<pass>@<dtu-ip>/api/dtu/capture"
# {"capture_mode":true}
```

### Files Changed (Capture Mode)

- `lib/Hoymiles/src/HoymilesRadio_CMT.cpp` — Capture mode: channel hopping, frame logging, CRC error reporting
- `lib/Hoymiles/src/HoymilesRadio_CMT.h` — Capture mode state and configuration
- `src/WebApi_dtu.cpp` — REST API endpoints (`/api/dtu/capture` GET/POST)
- `include/WebApi_dtu.h` — Handler declarations
- `docs/CaptureMode.md` — Full documentation

## Testing

Tested on:
- **Hardware**: OpenDTU Fusion v2 board (ESP32-S3)
- **Inverter**: MIT-5000-8T (SN: 1520a025566b) with 8 panels across 4 MPPT inputs
- **Conditions**: Low power (~230W, winter) and medium power (~1.8kW) captures
- **Validation**: Cross-referenced all fields against DTU-Pro Modbus TCP data via Home Assistant
- **Live polling**: MIT RealTimeRunData achieving ~100% success rate with retransmit tuning, SystemConfigPara parsing correctly at 38 bytes, all 4 MPPT channels + 3-phase AC data updating in web UI

### DIFF
--- a/docs/CaptureMode.md
+++ b/docs/CaptureMode.md
@@ -1,0 +1,120 @@
+# RF Capture Mode
+
+Capture mode enables passive RF sniffing on the CMT2300A radio, logging all received frames from Hoymiles inverters to the serial console (WebSocket). This is used for protocol reverse-engineering of unsupported inverter models.
+
+## Enabling Capture Mode
+
+### Via Web API
+
+```bash
+# Enable capture mode
+curl -X POST "http://admin:<password>@<dtu-ip>/api/dtu/capture" \
+  --data-urlencode 'data={"capture_mode":true}'
+
+# Check capture mode status
+curl -s "http://admin:<password>@<dtu-ip>/api/dtu/capture"
+# Returns: {"capture_mode":true}
+
+# Disable capture mode
+curl -X POST "http://admin:<password>@<dtu-ip>/api/dtu/capture" \
+  --data-urlencode 'data={"capture_mode":false}'
+```
+
+**Important:** The API uses form-encoded POST with a `data` parameter containing JSON. This is the standard OpenDTU `parseRequestData()` pattern — do **not** send a raw JSON body with `Content-Type: application/json`.
+
+### API Endpoints
+
+| Method | Endpoint | Description |
+|--------|----------|-------------|
+| GET | `/api/dtu/capture` | Returns current capture mode state |
+| POST | `/api/dtu/capture` | Enable/disable capture mode |
+
+POST body (form-encoded):
+```
+data={"capture_mode":true}
+```
+
+Both endpoints require HTTP Basic authentication (`admin:<password>`).
+
+## Reading Capture Output
+
+Capture frames are logged to the serial console, accessible via WebSocket:
+
+```
+ws://<dtu-ip>/console
+```
+
+### Python Example
+
+```python
+import websocket, re
+
+ws = websocket.create_connection("ws://<dtu-ip>/console", timeout=10)
+while True:
+    try:
+        ws.settimeout(3)
+        msg = ws.recv()
+        if 'CAPTURE' in msg and 'src=' in msg:
+            print(msg.strip())
+    except websocket.WebSocketTimeoutException:
+        continue
+ws.close()
+```
+
+### Log Format
+
+```
+I (timestamp) hoymiles: CAPTURE 863.00 MHz | -82 dBm | src=a025566b dst=a00e4ff1 len=27 | 95 A0 25 56 6B A0 0E 4F F1 01 ...
+```
+
+Fields:
+- **Frequency**: RF channel (863.00-870.00 MHz for EU)
+- **RSSI**: Signal strength in dBm
+- **src**: Source inverter serial (last 4 bytes)
+- **dst**: Destination DTU serial (last 4 bytes)
+- **len**: Total frame length in bytes
+- **Hex dump**: Complete RF frame
+
+### Frame Structure
+
+Each RF frame is 27 bytes:
+
+| Offset | Length | Field |
+|--------|--------|-------|
+| 0 | 1 | Preamble (0x95) |
+| 1-4 | 4 | Source serial (inverter) |
+| 5-8 | 4 | Destination serial (DTU) |
+| 9 | 1 | Frame ID / fragment count |
+| 10-25 | 16 | Payload data |
+| 26 | 1 | CRC8 (per-frame) |
+
+Frame IDs:
+- **0x01-0x05**: Inverter data response fragments
+- **0x80**: DTU data request
+- **0x81**: DTU ping/keepalive
+- **0x86**: Inverter final data fragment (shorter payload, includes CRC16)
+
+### Reassembly
+
+The OpenDTU framework reassembles response fragments:
+
+```c
+// InverterAbstract.cpp
+memcpy(_rxFragmentBuffer[fragmentId - 1].fragment, &fragment[10], len - 11);
+_rxFragmentBuffer[fragmentId - 1].len = len - 11;
+```
+
+This strips the 10-byte header and 1-byte CRC8, storing **16 data bytes per fragment**.
+
+## Channel Hopping
+
+When capture mode is active, the CMT radio sweeps through channels 12-40 (863.00-870.00 MHz in EU) to find active inverters. The hopping rate is automatic — it dwells on each channel briefly before moving to the next.
+
+Known active frequencies for the MIT-5000-8T: **863.00 MHz** and **863.25 MHz**.
+
+## Notes
+
+- Capture mode and normal inverter polling are **mutually exclusive**. While capture mode is active, the DTU will not poll configured inverters.
+- The WebSocket console has a **single-client limit**. If a browser tab has the console open, a Python client may receive no data.
+- Capture mode state does **not persist** across reboots. It must be re-enabled after each power cycle or firmware update.
+- Low-power conditions (winter, early morning) result in sporadic frame transmission. A complete 6-frame capture may take several minutes.
--- a/include/WebApi_dtu.h
+++ b/include/WebApi_dtu.h
@@ -12,6 +12,8 @@ public:
 private:
     void onDtuAdminGet(AsyncWebServerRequest* request);
     void onDtuAdminPost(AsyncWebServerRequest* request);
+    void onCaptureGet(AsyncWebServerRequest* request);
+    void onCapturePost(AsyncWebServerRequest* request);
 
     Task _applyDataTask;
     void applyDataTaskCb();
--- a/lib/CMT2300a/cmt2300a.c
+++ b/lib/CMT2300a/cmt2300a.c
@@ -82,13 +82,6 @@ bool CMT2300A_AutoSwitchStatus(uint8_t nGoCmd)
         break;
     }
 
-    /* When entering RX, set RX_AUTO_EXIT_DIS so the radio stays in RX
-     * after receiving a packet. Without this, back-to-back fragments
-     * in multi-packet responses are missed during the STBY gap. */
-    if (CMT2300A_GO_RX == nGoCmd) {
-        nGoCmd |= CMT2300A_MASK_RX_AUTO_EXIT_DIS;
-    }
-
     CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, nGoCmd);
 
     while (CMT2300A_GetTickCount() - nBegTick < 10) {
@@ -97,14 +90,14 @@ bool CMT2300A_AutoSwitchStatus(uint8_t nGoCmd)
         if (nWaitStatus == CMT2300A_GetChipStatus())
             return true;
 
-        if (CMT2300A_GO_TX == (nGoCmd & ~CMT2300A_MASK_RX_AUTO_EXIT_DIS)) {
+        if (CMT2300A_GO_TX == nGoCmd) {
             CMT2300A_DelayUs(100);
 
             if (CMT2300A_MASK_TX_DONE_FLG & CMT2300A_ReadReg(CMT2300A_CUS_INT_CLR1))
                 return true;
         }
 
-        if (CMT2300A_GO_RX == (nGoCmd & ~CMT2300A_MASK_RX_AUTO_EXIT_DIS)) {
+        if (CMT2300A_GO_RX == nGoCmd) {
             CMT2300A_DelayUs(100);
 
             if (CMT2300A_MASK_PKT_OK_FLG & CMT2300A_ReadReg(CMT2300A_CUS_INT_FLAG))
@@ -115,9 +108,6 @@ bool CMT2300A_AutoSwitchStatus(uint8_t nGoCmd)
     return false;
 
 #else
-    if (CMT2300A_GO_RX == nGoCmd) {
-        nGoCmd |= CMT2300A_MASK_RX_AUTO_EXIT_DIS;
-    }
     CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, nGoCmd);
     return true;
 #endif
--- a/lib/CMT2300a/cmt2300a.c
+++ b/lib/CMT2300a/cmt2300a.c
@@ -82,6 +82,13 @@ bool CMT2300A_AutoSwitchStatus(uint8_t nGoCmd)
         break;
     }
 
+    /* When entering RX, set RX_AUTO_EXIT_DIS so the radio stays in RX
+     * after receiving a packet. Without this, back-to-back fragments
+     * in multi-packet responses are missed during the STBY gap. */
+    if (CMT2300A_GO_RX == nGoCmd) {
+        nGoCmd |= CMT2300A_MASK_RX_AUTO_EXIT_DIS;
+    }
+
     CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, nGoCmd);
 
     while (CMT2300A_GetTickCount() - nBegTick < 10) {
@@ -90,14 +97,14 @@ bool CMT2300A_AutoSwitchStatus(uint8_t nGoCmd)
         if (nWaitStatus == CMT2300A_GetChipStatus())
             return true;
 
-        if (CMT2300A_GO_TX == nGoCmd) {
+        if (CMT2300A_GO_TX == (nGoCmd & ~CMT2300A_MASK_RX_AUTO_EXIT_DIS)) {
             CMT2300A_DelayUs(100);
 
             if (CMT2300A_MASK_TX_DONE_FLG & CMT2300A_ReadReg(CMT2300A_CUS_INT_CLR1))
                 return true;
         }
 
-        if (CMT2300A_GO_RX == nGoCmd) {
+        if (CMT2300A_GO_RX == (nGoCmd & ~CMT2300A_MASK_RX_AUTO_EXIT_DIS)) {
             CMT2300A_DelayUs(100);
 
             if (CMT2300A_MASK_PKT_OK_FLG & CMT2300A_ReadReg(CMT2300A_CUS_INT_FLAG))
@@ -108,6 +115,9 @@ bool CMT2300A_AutoSwitchStatus(uint8_t nGoCmd)
     return false;
 
 #else
+    if (CMT2300A_GO_RX == nGoCmd) {
+        nGoCmd |= CMT2300A_MASK_RX_AUTO_EXIT_DIS;
+    }
     CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, nGoCmd);
     return true;
 #endif
--- a/lib/CMT2300a/cmt2300wrapper.cpp
+++ b/lib/CMT2300a/cmt2300wrapper.cpp
@@ -35,13 +35,6 @@ bool CMT2300A::startListening(void)
     CMT2300A_EnableReadFifo();
     CMT2300A_ClearRxFifo();
 
-    // Disable RX auto-exit: keep the radio in RX mode after receiving a packet.
-    // Without this, the CMT2300A transitions to STBY after each packet, causing
-    // subsequent fragments in a multi-packet burst to be missed.
-    uint8_t modeCtl = CMT2300A_ReadReg(CMT2300A_CUS_MODE_CTL);
-    modeCtl |= CMT2300A_MASK_RX_AUTO_EXIT_DIS;
-    CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, modeCtl);
-
     if (!CMT2300A_GoRx()) {
         return false;
     } else {
--- a/lib/CMT2300a/cmt2300wrapper.cpp
+++ b/lib/CMT2300a/cmt2300wrapper.cpp
@@ -64,6 +64,13 @@ void CMT2300A::read(void* buf, const uint8_t len)
     CMT2300A_ReadFifo(static_cast<uint8_t*>(buf), len);
 
     CMT2300A_ClearInterruptFlags();
+
+    // After receiving a packet, the CMT2300A exits RX mode (goes to STBY).
+    // Re-enter RX as fast as possible to catch the next fragment in a burst.
+    // Skip the full startListening() sequence (GoStby + EnableReadFifo +
+    // ClearRxFifo + GoRx) — just write GO_RX directly to MODE_CTL.
+    // This is ~4x faster and catches more fragments in rapid bursts.
+    CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, CMT2300A_GO_RX);
 }
 
 bool CMT2300A::write(const uint8_t* buf, const uint8_t len)
--- a/lib/CMT2300a/cmt2300wrapper.cpp
+++ b/lib/CMT2300a/cmt2300wrapper.cpp
@@ -35,6 +35,13 @@ bool CMT2300A::startListening(void)
     CMT2300A_EnableReadFifo();
     CMT2300A_ClearRxFifo();
 
+    // Disable RX auto-exit: keep the radio in RX mode after receiving a packet.
+    // Without this, the CMT2300A transitions to STBY after each packet, causing
+    // subsequent fragments in a multi-packet burst to be missed.
+    uint8_t modeCtl = CMT2300A_ReadReg(CMT2300A_CUS_MODE_CTL);
+    modeCtl |= CMT2300A_MASK_RX_AUTO_EXIT_DIS;
+    CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, modeCtl);
+
     if (!CMT2300A_GoRx()) {
         return false;
     } else {
@@ -64,13 +71,6 @@ void CMT2300A::read(void* buf, const uint8_t len)
     CMT2300A_ReadFifo(static_cast<uint8_t*>(buf), len);
 
     CMT2300A_ClearInterruptFlags();
-
-    // The CMT2300A auto-exits RX mode after receiving a packet (goes to STBY).
-    // Re-enter RX immediately so back-to-back fragments from MIT inverters
-    // (which arrive at ~50ms intervals within a burst) are not missed.
-    CMT2300A_EnableReadFifo();
-    CMT2300A_ClearRxFifo();
-    CMT2300A_GoRx();
 }
 
 bool CMT2300A::write(const uint8_t* buf, const uint8_t len)
--- a/lib/CMT2300a/cmt2300wrapper.cpp
+++ b/lib/CMT2300a/cmt2300wrapper.cpp
@@ -64,6 +64,13 @@ void CMT2300A::read(void* buf, const uint8_t len)
     CMT2300A_ReadFifo(static_cast<uint8_t*>(buf), len);
 
     CMT2300A_ClearInterruptFlags();
+
+    // The CMT2300A auto-exits RX mode after receiving a packet (goes to STBY).
+    // Re-enter RX immediately so back-to-back fragments from MIT inverters
+    // (which arrive at ~50ms intervals within a burst) are not missed.
+    CMT2300A_EnableReadFifo();
+    CMT2300A_ClearRxFifo();
+    CMT2300A_GoRx();
 }
 
 bool CMT2300A::write(const uint8_t* buf, const uint8_t len)
--- a/lib/CMT2300a/cmt2300wrapper.cpp
+++ b/lib/CMT2300a/cmt2300wrapper.cpp
@@ -64,15 +64,6 @@ void CMT2300A::read(void* buf, const uint8_t len)
     CMT2300A_ReadFifo(static_cast<uint8_t*>(buf), len);
 
     CMT2300A_ClearInterruptFlags();
-
-    // Re-enter RX mode so the next fragment in a burst can be received.
-    // The CMT2300A exits RX after each packet (goes to STBY).
-    // Use the full startListening() sequence to properly reset the FIFO
-    // and re-enter RX — a bare GoRx write corrupts FIFO reads.
-    CMT2300A_GoStby();
-    CMT2300A_EnableReadFifo();
-    CMT2300A_ClearRxFifo();
-    CMT2300A_GoRx();
 }
 
 bool CMT2300A::write(const uint8_t* buf, const uint8_t len)
--- a/lib/CMT2300a/cmt2300wrapper.cpp
+++ b/lib/CMT2300a/cmt2300wrapper.cpp
@@ -65,12 +65,14 @@ void CMT2300A::read(void* buf, const uint8_t len)
 
     CMT2300A_ClearInterruptFlags();
 
-    // After receiving a packet, the CMT2300A exits RX mode (goes to STBY).
-    // Re-enter RX as fast as possible to catch the next fragment in a burst.
-    // Skip the full startListening() sequence (GoStby + EnableReadFifo +
-    // ClearRxFifo + GoRx) — just write GO_RX directly to MODE_CTL.
-    // This is ~4x faster and catches more fragments in rapid bursts.
-    CMT2300A_WriteReg(CMT2300A_CUS_MODE_CTL, CMT2300A_GO_RX);
+    // Re-enter RX mode so the next fragment in a burst can be received.
+    // The CMT2300A exits RX after each packet (goes to STBY).
+    // Use the full startListening() sequence to properly reset the FIFO
+    // and re-enter RX — a bare GoRx write corrupts FIFO reads.
+    CMT2300A_GoStby();
+    CMT2300A_EnableReadFifo();
+    CMT2300A_ClearRxFifo();
+    CMT2300A_GoRx();
 }
 
 bool CMT2300A::write(const uint8_t* buf, const uint8_t len)
--- a/lib/Hoymiles/src/Hoymiles.cpp
+++ b/lib/Hoymiles/src/Hoymiles.cpp
@@ -13,6 +13,7 @@
 #include "inverters/HMS_4CH.h"
 #include "inverters/HMT_4CH.h"
 #include "inverters/HMT_6CH.h"
+#include "inverters/MIT_8CH.h"
 #include "inverters/HM_1CH.h"
 #include "inverters/HM_2CH.h"
 #include "inverters/HM_4CH.h"
@@ -153,7 +154,9 @@ void HoymilesClass::loop()
 std::shared_ptr<InverterAbstract> HoymilesClass::addInverter(const char* name, const uint64_t serial)
 {
     std::shared_ptr<InverterAbstract> i = nullptr;
-    if (HMT_4CH::isValidSerial(serial)) {
+    if (MIT_8CH::isValidSerial(serial)) {
+        i = std::make_shared<MIT_8CH>(_radioCmt.get(), serial);
+    } else if (HMT_4CH::isValidSerial(serial)) {
         i = std::make_shared<HMT_4CH>(_radioCmt.get(), serial);
     } else if (HMT_6CH::isValidSerial(serial)) {
         i = std::make_shared<HMT_6CH>(_radioCmt.get(), serial);
--- a/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
+++ b/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
@@ -173,7 +173,7 @@ void HoymilesRadio_CMT::loop()
         uint8_t drainCount = 0;
         uint32_t lastRxTime = millis();
         while (true) {
-            if (_radio->available()) {
+            if (_radio->available() || _radio->rxFifoAvailable()) {
                 if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
                     ESP_LOGE(TAG, "CMT2300A: Buffer full");
                     _radio->flush_rx();
--- a/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
+++ b/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
@@ -166,30 +166,35 @@ void HoymilesRadio_CMT::loop()
     }
 
     // Step 1: Drain all available packets from the hardware FIFO into the
-    // software ring buffer. This runs every iteration — not just when the
-    // interrupt flag is set — so that back-to-back packets from MIT inverters
-    // (which send 6 rapid response fragments) are captured before the 64-byte
-    // hardware FIFO overflows.
+    // software ring buffer. After receiving a packet, poll for additional
+    // packets for up to 80ms to catch back-to-back fragments in a burst
+    // (fragments arrive at ~50ms intervals from HMS/HMT/MIT inverters).
     if (_packetReceived) {
         uint8_t drainCount = 0;
-        const uint32_t drainStart = millis();
-        while (_radio->available()) {
-            if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
-                ESP_LOGE(TAG, "CMT2300A: Buffer full");
-                _radio->flush_rx();
+        uint32_t lastRxTime = millis();
+        while (true) {
+            if (_radio->available()) {
+                if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
+                    ESP_LOGE(TAG, "CMT2300A: Buffer full");
+                    _radio->flush_rx();
+                    break;
+                }
+
+                fragment_t f;
+                memset(f.fragment, 0xcc, MAX_RF_PAYLOAD_SIZE);
+                f.len = std::min<uint8_t>(_radio->getDynamicPayloadSize(), MAX_RF_PAYLOAD_SIZE);
+                f.channel = _radio->getChannel();
+                f.rssi = _radio->getRssiDBm();
+                f.wasReceived = false;
+                f.mainCmd = 0x00;
+                _radio->read(f.fragment, f.len);
+                _rxBuffer.push(f);
+                drainCount++;
+                lastRxTime = millis();
+            } else if (millis() - lastRxTime > 80) {
+                // No new packet for 80ms — burst is over
                 break;
             }
-
-            fragment_t f;
-            memset(f.fragment, 0xcc, MAX_RF_PAYLOAD_SIZE);
-            f.len = std::min<uint8_t>(_radio->getDynamicPayloadSize(), MAX_RF_PAYLOAD_SIZE);
-            f.channel = _radio->getChannel();
-            f.rssi = _radio->getRssiDBm();
-            f.wasReceived = false;
-            f.mainCmd = 0x00;
-            _radio->read(f.fragment, f.len);
-            _rxBuffer.push(f);
-            drainCount++;
         }
         _radio->flush_rx();
         _packetReceived = false;
--- a/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
+++ b/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
@@ -171,7 +171,8 @@ void HoymilesRadio_CMT::loop()
     // (which send 6 rapid response fragments) are captured before the 64-byte
     // hardware FIFO overflows.
     if (_packetReceived) {
-        ESP_LOGV(TAG, "Interrupt received");
+        uint8_t drainCount = 0;
+        const uint32_t drainStart = millis();
         while (_radio->available()) {
             if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
                 ESP_LOGE(TAG, "CMT2300A: Buffer full");
@@ -188,6 +189,7 @@ void HoymilesRadio_CMT::loop()
             f.mainCmd = 0x00;
             _radio->read(f.fragment, f.len);
             _rxBuffer.push(f);
+            drainCount++;
         }
         _radio->flush_rx();
         _packetReceived = false;
--- a/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
+++ b/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
@@ -166,35 +166,24 @@ void HoymilesRadio_CMT::loop()
     }
 
     // Step 1: Drain all available packets from the hardware FIFO into the
-    // software ring buffer. After receiving a packet, poll for additional
-    // packets for up to 80ms to catch back-to-back fragments in a burst
-    // (fragments arrive at ~50ms intervals from HMS/HMT/MIT inverters).
+    // software ring buffer.
     if (_packetReceived) {
-        uint8_t drainCount = 0;
-        uint32_t lastRxTime = millis();
-        while (true) {
-            if (_radio->available() || _radio->rxFifoAvailable()) {
-                if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
-                    ESP_LOGE(TAG, "CMT2300A: Buffer full");
-                    _radio->flush_rx();
-                    break;
-                }
-
-                fragment_t f;
-                memset(f.fragment, 0xcc, MAX_RF_PAYLOAD_SIZE);
-                f.len = std::min<uint8_t>(_radio->getDynamicPayloadSize(), MAX_RF_PAYLOAD_SIZE);
-                f.channel = _radio->getChannel();
-                f.rssi = _radio->getRssiDBm();
-                f.wasReceived = false;
-                f.mainCmd = 0x00;
-                _radio->read(f.fragment, f.len);
-                _rxBuffer.push(f);
-                drainCount++;
-                lastRxTime = millis();
-            } else if (millis() - lastRxTime > 80) {
-                // No new packet for 80ms — burst is over
+        while (_radio->available()) {
+            if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
+                ESP_LOGE(TAG, "CMT2300A: Buffer full");
+                _radio->flush_rx();
                 break;
             }
+
+            fragment_t f;
+            memset(f.fragment, 0xcc, MAX_RF_PAYLOAD_SIZE);
+            f.len = std::min<uint8_t>(_radio->getDynamicPayloadSize(), MAX_RF_PAYLOAD_SIZE);
+            f.channel = _radio->getChannel();
+            f.rssi = _radio->getRssiDBm();
+            f.wasReceived = false;
+            f.mainCmd = 0x00;
+            _radio->read(f.fragment, f.len);
+            _rxBuffer.push(f);
         }
         _radio->flush_rx();
         _packetReceived = false;
--- a/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
+++ b/lib/Hoymiles/src/HoymilesRadio_CMT.cpp
@@ -116,6 +116,10 @@ void HoymilesRadio_CMT::init(const int8_t pin_sdio, const int8_t pin_clk, const 
     }
 
     _isInitialized = true;
+
+    // Start in RX mode so passive capture works without needing to TX first
+    _radio->startListening();
+    ESP_LOGI(TAG, "CMT2300A: RX mode active");
 }
 
 void HoymilesRadio_CMT::loop()
@@ -124,19 +128,55 @@ void HoymilesRadio_CMT::loop()
         return;
     }
 
+    // Capture mode: hop across all legal EU channels to find traffic
+    if (_captureMode && !_busyFlag) {
+        const uint32_t now = millis();
+        if (now - _captureLastHop >= CAPTURE_HOP_INTERVAL_MS) {
+            _captureLastHop = now;
+
+            // Calculate channel range from legal frequency limits
+            const uint8_t minCh = getChannelFromFrequency(countryDefinition.at(_countryMode).Freq_Legal_Min);
+            const uint8_t maxCh = getChannelFromFrequency(countryDefinition.at(_countryMode).Freq_Legal_Max);
+
+            if (minCh != 0xFF && maxCh != 0xFF) {
+                _captureChIdx++;
+                if (_captureChIdx > maxCh || _captureChIdx < minCh) {
+                    _captureChIdx = minCh;
+                }
+                // Switch channel while staying in RX — GoStby, change channel, GoRx
+                _radio->stopListening();
+                _radio->setChannel(_captureChIdx);
+                _radio->startListening();
+
+                // Log channel sweep start for diagnostics
+                if (_captureChIdx == minCh) {
+                    ESP_LOGD(TAG, "CAPTURE: sweep restart ch %" PRIu8 "-%" PRIu8 " (%.2f-%.2f MHz)",
+                        minCh, maxCh,
+                        getFrequencyFromChannel(minCh) / 1000000.0,
+                        getFrequencyFromChannel(maxCh) / 1000000.0);
+                }
+            }
+        }
+    }
+
     if (!_gpio3_configured) {
         if (_radio->rxFifoAvailable()) { // read INT2, PKT_OK flag
             _packetReceived = true;
         }
     }
 
+    // Step 1: Drain all available packets from the hardware FIFO into the
+    // software ring buffer. This runs every iteration — not just when the
+    // interrupt flag is set — so that back-to-back packets from MIT inverters
+    // (which send 6 rapid response fragments) are captured before the 64-byte
+    // hardware FIFO overflows.
     if (_packetReceived) {
         ESP_LOGV(TAG, "Interrupt received");
         while (_radio->available()) {
             if (_rxBuffer.size() > FRAGMENT_BUFFER_SIZE) {
                 ESP_LOGE(TAG, "CMT2300A: Buffer full");
                 _radio->flush_rx();
-                continue;
+                break;
             }
 
             fragment_t f;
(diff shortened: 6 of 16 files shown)


tbnobody:
Thank you!
 In #2742 you wrote, that the main loop is too slow to handle the received packages (reading it from the CMT module). I just had a look at your code and didn't find any changes related to this. Is there still any improvement required to get this inverter working?

Geoffn-Hub:
I'm going to submit a dedicated PR for the changes to the receive loop - I figured you might want to have the parsing logic for the MIT inverter packaged seperately from code that alters the core code of the project in more fundamental ways

Geoffn-Hub:
FWIW - I did take a look at the AhoyDTU implementation which is in the form of a state machine, but don't think it works in this case either

Geoffn-Hub (commented):
Good point about longer sequences — worth thinking ahead. A few observations from digging deeper:

The ring buffer drain is fast (microsecond-level per fragment), so 15-16 packets aren't a problem from a buffering perspective.

The bigger factor for MIT is per-packet frequency hopping. Looking at the CMT2300A datasheet, this is actually a designed hardware feature — "fast manual frequency hopping" via `FH_CHANNEL`/`FH_OFFSET` registers, with a dedicated app note ([AN197](https://www.hoperf.com/uploads/AN197-CMT2300A-CMT2119B-CMT2219B_Fast_Manual_Frequemcy_Hopping_EN_V0.8_1757313080.pdf)) covering RX-side hopping including AFC reconfiguration.

OpenDTU already sets `FH_OFFSET=100` (250 kHz steps) during init — which matches exactly what we observed on the MIT. The current RX hop implementation in this PR uses the heavyweight `stopListening→setChannel→startListening` cycle. We can optimise this to a single `FH_CHANNEL` register write, which is what the hardware FH feature is designed for.

We also need to add `CMT2300A_SetAfcOvfTh()` per AN197 for proper RX-side hopping — without it the AFC circuit may not be properly calibrated after channel switches.

For HMS/HMT, longer sequences should work fine on a single channel — the per-packet hopping appears to be MIT-specific behaviour (HMS/HMT consistently deliver all fragments on the same frequency). But having the FH infrastructure in place would also benefit the frequency-drift issue (#2137) down the road.

stefan123t:
Adding @lumapu @DanielR92 to the party :tada: as they have done the initial research and CMT implementation for HMS. 
Maybe they can answer or add something to the CMT2300A frequency hopping ?

Found the documentation you mentioned under **2. AFC Parameter Setting for RX Frequency Hopping**:

<details closed><summary>Details</summary>
<p>

> When using manual frequency hopping in RX, the frequency hopping operation will affect the setting of an
important parameter `AFC_OVF_TH<7:0>` of **AFC** (*Automatic Frequency Control*). <ins>If the setting is incorrect, it may
result in failure to receive at the current frequency.</ins> Under normal circumstances, this parameter is automatically
calculated by RFPDK based on the frequency points input by the user, and the user does not need to configure
it specially. However, when users use **fast frequency hopping** or manually configure frequency points, the
setting of frequency points is separated from RFPDK, so RFPDK cannot help users configure this parameter.
CMOSTEK provides another tool for customers to automatically calculate this parameter, which is the
"CMT2300A-CMT2219B Frequency Hopping Calculation Table". The usage method is as follows:
> 1. The user fills in 4 parameters respectively.
> <ins>Xtal Frequency (MHz)</ins> - Crystal frequency, the default is 26MHz. If necessary, modify <ins>RX Xtal Tolerance (ppm)</ins>
according to the actual frequency - The maximum error of the crystal used by the receiver
<ins>TX Xtal Tolerance (ppm)</ins> - Crystal error used by the transmitter <ins>RX Frequency Range (MHz)</ins> - Frequency
hopping range of the receiver
Note that the crystal errors related to TX and RX refer to the crystal errors of the respective communication
parties. For TRX products, one party can be assumed to be RX and the other to be TX.
> 2. Then, in one column of the Index, fill in each frequency point of frequency hopping that needs to be used
in order, with a maximum of 100 entries. If the user does not use that many frequency points, the
remaining ones can be left blank.
> 3. After filling in all the frequency hopping points, the table will calculate the values (in decimal) of
`AFC_OVF_TH<7:0>` corresponding to each frequency point. At the same time, the value of Initial
`AFC_OVF_TH<7:0>` will be displayed in the left table.
> * If **all** values of `AFC_OVF_TH<7:0>` are **equivalent to Initial `AFC_OVF_TH<7:0>`**, it indicates that there
are no frequency points requiring special handling. In this case, the user only needs to **write** the value of
**Initial `AFC_OVF_TH<7:0>`** into the `0x27 CUS_FSK4` register **once** during the chip initialization process,
and there will be no need to make changes afterward.
> * If in a certain frequency point, the value of `AFC_OVF_TH<7:0>` is **different from that of Initial
`AFC_OVF_TH<7:0>`**, then this value will be marked in red. The user needs to configure the value of Initial
`AFC_OVF_ TH<7:0>` during the initialization process. **Before hopping** to this frequency point (in the `STBY`
state), **update the corresponding value** of `AFC_OVF_TH<7:0>` to the `0x27 CUS_FSK4` register. Before
jumping to the next frequency point, if the two values are the same, it is necessary to restore the register
content to the value of Initial `AFC_OVF_TH<7:0>`.
> 
> In short, users need to ensure that before jumping to each frequency point, the value of the `0x27 CUS_FSK`
register must be the same as the corresponding `AFC_OVF_TH<7:0>` value in the table

</p>
</details> 

and **3. The Overall Process of Rapid Manual Frequency Hopping**

> The following summarizes the overall process of fast manual frequency hopping:
> 1. Use the CMT2300A-CMT2219B Frequency Hopping Calculation Table to obtain the **Initial `AFC_OVF_TH<7:0>`**, and write this value into the `0x27 CUSFSK4` register during the chip initialization process. 
>   *If only TX frequency hopping is used, this step can be ignored.*
> 2. **Send `go_stby`** to make the chip return to the `STBY` state.
> 3. **Set `FH CHANNEL<7:0>`**. 
>   If it is necessary to reset `FH_OFFSET<7:0>`, handle it in this step as well. 
>   [...] 
> <details closed><summary>show fine print</summary>
>   <p>
> 
>   Refer to the operating frequency band division in Table 3 to limit the frequency hopping range.
>   It should be kept in mind that the chip does not support cross-band frequency hopping operations, that is, *the value of the `FREQ_DIVX_CODE<2:0>` register cannot be changed*.
>   Since the base frequency is determined by RFPDK, the parameter table generated by default by RFPDK includes the value of the `FREQ_DIVX_CODE<2:0>` register corresponding to this frequency band, which is configured during initialization, so users do *not need to configure it every time* they *frequency hop*.
>   However, it should be noted that `FREQ_VCO_BANK<2:0>` will change within the same frequency band.
>   When users set the frequency hopping points in this step, corresponding to the range marked in Table 3, if necessary, they should update the value of `FREQ_VCO_BANK<2:0>`.
> 
>   </p>
> </details> 
> 
> 4. If performing **RX frequency hopping** operation, according to the method introduced in Chapter 2, use the EXCEL tool sheet provided by CMOSTEK to **ensure** that the value of the **`0x27 CUS_FSK4` register is equal to the value of `AFC_OVF_TH<7:0>`** corresponding to the current frequency point to be hopped to.
>   *If it is a TX frequency hopping operation, this step can be ignored.*
> 5. **Send `go_tx`** to enter **TX for transmission**, or **`go_rx`** to enter **RX for reception**.
> 
> If the user performs a frequency hopping operation on TX and then **switches to RX** state for receiving, they
**must set `FH_CHANNEL<7:0>` back to `0`** before reception to return to the RX basic frequency point set by
RFPDK. Conversely, **after** performing the **frequency hopping** operation of RX, **`FH_CHANNEL<7:0>` must also
be set back to `0`** to using the **TX** basic frequency point for transmission, setting by RFPDK.
> 
> The advantage of manual frequency hopping is that the frequency point can be changed by setting only 1 to 2
registers. Without this mechanism, it would be necessary to write to the frequency region to change the
frequency, which would significantly increase the content of the registers that need to be written, lengthen the
time, and might fail to meet application requirements.

Geoffn-Hub:
Happy for help.  :-)

Other than the buffer overflow issue - this is the main problem that was preventing us from getting a full capture from the MIT inverter, and I don't have it fully beaten yet - I have to go through several retransmit request cycles for complete updates to succeed.

image

Geoffn-Hub:
To save hunting back through the comments...this is what we are seeing from the captures

CAPTURE 863.50 MHz | frag 01 (src=a025566b)
CAPTURE 863.75 MHz | frag 02
CAPTURE 864.00 MHz | frag 03
...
CAPTURE 863.50 MHz | frag 04
CAPTURE 863.75 MHz | frag 05


The MIT *IS hopping across 3 frequencies*: 863.50, 863.75, 864.00 MHz —  250 kHz steps. The pattern per fragment ID:
•⁠  ⁠Frag 1 → 863.50 (-1)
•⁠  ⁠Frag 2 → 863.75 (base)
•⁠  ⁠Frag 3 → 864.00 (+1)
•⁠  ⁠Frag 4 → 863.50 (-1) ← repeats
•⁠  ⁠Frag 5 → 863.75 (base)

stefan123t:
@Geoffn-Hub any news from your adventures in MIT Frequency Hopping ? 
I have been on holidays so did not check back regularly.

Dirk-Ahoy:
@stefan123t  -  es könnte sein das HM das auch in die neuere Chargen der HM1600 eingebaut hat   S/N 1164A02...   

siehe Discord Opendtu Channel, wir haben geloggt - da werden nur ein paar Frames empfangen - kannst dich ja mal im Discord melden  dann sende ich dir die Logs.    Ist aber alles nur Vermutung im Moment.     An der Firmware kann es nicht liegen, wir haben da auf dem WR ein Downgrade auf 1.1.12 veranlasst -  die sollten funktioniern - fragt sich ob die im Funkmodul was an der Firmware oder Timings geschraubt haben.

Zurrmaxe:
Hallo zusammen 
Ich habe mir jetzt auch ein paar MIT-5000 zugelegt und bin auf der Suche nach eine Möglichkeit, erstmal nur die wichtigtsten Daten (z.B. die aktuelle Leistung) für weitere eine Verarbeitungen auslesen zu können. 
Dann könnte ich auch mitlesen und testen, wie der Datenfluss und die Kommandos eventuell funktionieren.

Gibt es von eueren Versuchen schon eine BETA-Version ?   

lg Alf

stefan123t:
@Zurrmaxe Du bist im PR von Geoffn-Hub da sind alle Änderungen drin. Einfach clonen, ggf den Branch wechseln und den Code bauen...

Zurrmaxe:
@stefan123t 
Ich bin leider nur mit der Arduino IDE unterwegs und hatte mich bisher noch nicht mit anderen Systemen wie Platformio pp. beschäftigt. 
Werde es aber heute abend mal versuchen. Anleitungen gibts bestimmt hier irgendwo im Netz.

LG
Alf

Zurrmaxe:
@stefan123t

Ich habe einen nagelneuen MIT5000 diese Woche angeschlossen. 
Brunch von Geoffn-Hub auf ESP32 gespielt und zusammen mit 2 x  HMT2250 getestet. 
Die HMT2250 werden ausgelesen, MIT5000 jedoch garnichts.

Nach Ostern gehts weiter...
Alf

rauschgiftengel:
> @Zurrmaxe Du bist im PR von Geoffn-Hub da sind alle Änderungen drin. Einfach clonen, ggf den Branch wechseln und den Code bauen...

Hallo Stefan,

ich hab das ebenfalls getestet, ich bekomme auch Werte vom MIT5000 auf der openDTU - allerdings funktioniert dann die DTU S Pro nicht mehr - auch wenn ich der OpenDTU die selbe ID gebe - ist das normal ? 
Ich meine irgendwo gelesen zu haben dass jemand die Werte der OpenDTU mit der DTU S Pro verglichen hat - in dem Fall müssten ja beide parallel funktioniert haben. 

@Zurrmaxe 

mit diesen Settings funktioniert es bei mir (branch capture_mode):
image

Zurrmaxe:
@rauschgiftengel
Bei mir läuft der Brunch (branch capture_mode) auch seit Tagen, nachdem ich einige Änderungen an der Hardware gemacht  habe. 
Folgendes habe ich festgestellt: 
Die Werte in der OpenDtu sind für mich ausreichend genau. 
Die Inverter starten in der Regel bei 868MHz und schalten auf Anforderung dann auf einen "freien" Kanal um.  

Der parallele Betrieb von zwei  DTUs gleichzeitig führt zu einer erhöhten Fehlerquote, weil die erste DTU die Frequenz   verschiebt und die andere DTU dann ins "Leere" funkt usw.
Hierdurch erhöht sich die Fehlerquote bis eine DTU aufgibt.

Wird ein "Mischbetrieb" von mehreren Invertern (HM800/HMT2250/MIT5000) mit einer (branch capture_mode)-OpenDTU gemacht, liegt die Trefferquote erkannter Pakete beim MIT5000 unter 10%,  obwohl sich der MIT5000 bei mir direkt neben der DTU befindet (-22db). Der Grund dürften in den unterschedlichen Übertragungsverfahren/Datentelegrammen  und die anschließende Auswertung dieser Pakete liegen.

Der bisherige Weg von Geoffn-Hub, tbnobody , stefan123tpp  ist schon sehr gut und sollte weiter verfolgt werden.

Eine "Leistungsregulierung" oder Nulleinspeisung pp. dürfte beim MIT5000 überflüssig sein und sollte nicht weiter verfolgt werden.
Ich kann mir nicht vorstellen, dass sich jemand einen MIT5000 an die Wand nagelt und ihn anschließend als 800W Balkonkraftwerk verkaufen will. 

lg
Alf.

Redzone003:
@Zurrmaxe , can you please help me with a compiled version for MIT5000? I'm referring to the bin file. I have an identical inverter and it's connected from today to a deye hybrid inverter and it keeps powering up and down, I can't figure out where the problem is - that's why I need to see what's happening with MIT5000, what the voltages are and if it's somehow limited.

Zurrmaxe:
@Redzone003 
Ssorry, but I have no ready "bin"-File for a ESP32.  

I worked with Visual Studio ,Plattformio and Arduino IDE to compile the brunch-code to my ESP32.
My ESP32 is connected to my computer and i use only the upload-function from Plattformio.  

You should do the same way. 

There are lots of sites on YouTube about "build and upload" ESP-Code with Plattformio and VS.

Alf

Zurrmaxe:
**Hallo zusammen**

Hier eine kleine Info zu meinen Versuchen in den letzten Wochen mit MIT5000

Ich habe seit einigen Wochen den MIT5000 mit der.o.a. Software am laufen und hatte keine Probleme mit dem MIT5000.

Allerdings musste ich feststellen, dass der MIT5000, im Gegensatz zu den anderen Wechselrichtern, erst später mit der Solareinspeisung beginnt. 

Während meine HMT2250 bereits ab 05.30 U(hr  arbeiten,  bleibt der MIT5000 noch aus. 
Erst Stunden  später fängt der MIT5000 dann an, Strom  zu produzieren.




Also ging ich an die Suche nach den Gründen 

Die Ursache liegt an den von mir verwendeten Sorlarplatten von JA Solar Mono mit 405W (108 Zellen). 
Die Solarplatten der HMT2250 und des MIT5000 sind zusammen in gleicher Richtung und Winkel montiert.

Wöhrend die MPPTs der HMT2250 eine Einschaltspannung von ca. 24 Volt benötigen, braucht der MIt5000 jedoch ungefähr 32 Volt (laut Datenblatt.)  Die von mir verwendeten JA-Solarmodule haben jedoch nur eine Spannung von 31,7 Volt (Laut Datenblatt unter Last).

Am jedem Eingang des  MIT5000 müssen immer 2 Platten angeschlossen werden.  

Jetzt kommen wir zu einer blödsinnigen Anzeige der Solarplatten in der originalen Hoymiles-DTU: 

Platten

 Wer eine Hoymiles-DTU Pro-2 zur Verfügung hat stellt fest, dass Hoymiles in der DTU-Software pro Eingang immer zwei Platten anzeigt und nicht die tatsächlichen Werte der Solarplatten, sondern nur die Gesamtsumme auf den zwei Platten **aufteilt.**   

Soalr


Bei den jetzigen hohen Aussentemperatiren sinkt die Spanung jedoch unter 32 Volt. (nachgemessen mit Multimeter).
Somit wird die MPPT-Einschaltspannung des MIT5000 nicht (immer) oder später erreicht.

Gleiches passierte mir auch mit anderen Solarplattem der 400W Baureihen.

Hat man bereits Solarplatten auf dem Dach, muss man mit dem Datenbaltt vergleichen, ob die erforderlichen Spannungen erreicht werden können.


Nachtrag vom 4.6.2026

Auf Nachfrage wurde mir von Hoymiles bestätigt, dass die MPPTs erst mit 32 Volt starten. 
Danach soll der Arbeitsbereich später zwischen 12 und 136 V liegen und somit auch bei niedrigen Spannungen funktionieren. Ein Update zur Reduzierung der Einschaltsspannung gibt es nicht.



**Mein Fazit: Die Anschaffuung eines  MIT5000 lohnt sich nur, wenn man auch enstsprechend große Solarmodule (ab 450W) verwenden will.** 

lg Alf

tronvers:
Hallo,

Würde der  MIT5000 denn am Akku laufen ?

Zurrmaxe:
> Hallo,
> 
> Würde der MIT5000 denn am Akku laufen ?

Zunächst ist festzustellen, dass Hoymiles MIT5000 4 Eingänge hat. 

An jedem Eingang müssen 2 Panels **in Reihe** angeschlossen werden um die Einschaltspannung von 64 Volt (2 x 32V) zu erreichen. 

So weit so gut. Doch jetzt wird es sehr komisch, wenn nicht sogar blödsinnig: 

In der Hoymiles-Software DTU-Pro werden, aber 8 Panels angezeigt (Siehe Bild oben, an nur 4 Eingängen ?!)

Da man aber die tatsächlichen Spannungen und Ströme der einzelnen Platten in einer **Reihenschaltung** gar nicht messen kann, teilt man die gemessenen Wert einfach durch zwei und auf die Platten auf. (siehe ebenfalls oben)

Somit geben die in der Hoymiles-DTU.Pro bzw. in Firmware des MIT50000 anzeigten Werte niemals die tatsächliche Spannung oder Leistung einer Platte da.

Man kann also nie feststellen, welche Platte welche Spannungen oder Leistungen tatsächlich erzeugt und Rückschlüsse auf deren Funktion machen. 

Wichtig ist, dass der MPTT-Eingang mehr als 64 Volt bekommt ( 2x32Volt ). 

Wenn also z.B. ein Speicher zwischen Solarpanel und MIT5000 angeschlossen wird, müssen die zwei Ausgänge eines  Speichers (z.B MarstekB2500)  ebenfalls in Reihe zusammengeschlossen werden um mehr als 64 Volt (zusammen) erzeugen
zu können.

Dazu ist das Datenplatt des jeweiligen Speichers genau zu lesen um festzustellen, ob es möglich so möglich ist.

Ein pauschale Aussage kann man nicht leider machen.

lg ALf

tronvers:
Unter dem letzten Bild von Zurrmaxe steht aber:

Auf Nachfrage wurde mir von Hoymiles bestätigt, dass die MPPTs erst mit 32 Volt starten.
Danach soll der Arbeitsbereich später zwischen 12 und 136 V liegen und somit auch bei niedrigen Spannungen funktionieren. Ein Update zur Reduzierung der Einschaltsspannung gibt es nicht.

Also könnte der MIT 5000 an einem 48 Volt AKKU arbeiten ......

Zurrmaxe:
> Arbeitsbereich später zwischen 12 und 136 V

Ja, aber **starten** zuerst mit 32(64) Volt !

fpoisson2:
**Tested on 3× HMS-2000-4T (North America / 915 MHz) - confirms the fix, with a hardware-revision data point**

I can confirm this PR resolves the "receive partial" communication failure also described in #3043, and my setup adds an interesting detail: all three of my inverters are HMS-2000-4T running the **exact same inverter firmware** (build 10100, 2024-10-16, bootloader 102), but they differ in hardware revision:

| Inverter | HW version | Stable v26.3.30 | With this PR |
|---|---|---|---|
| HMS-2000-4T | 02.00 | ~3% RX success, almost all `Retransmit timeout` | ✅ reachable, data flowing |
| HMS-2000-4T | 02.10 | ✅ 100% RX success | ✅ 100% RX success |
| HMS-2000-4T | 02.00 | ~10% RX success, mostly partial | ✅ reachable, data flowing |

So at least in my case the trigger is the **hardware revision (02.00 vs 02.10), not the inverter firmware version** — which may explain why same-model/same-firmware setups behave so differently for different users.

Symptoms on stable, for the record: strong RSSI (-46 to -60 dBm), first-pass fragments received, but retransmit requests had to be repeated ~3× before the inverter answered, exhausting the retransmit budget nearly every cycle (`RX Fail: Receive Partial`). The 02.00 units never completed a `DevInfo` request, so they stayed "unknown" in the UI despite producing.

Remaining with this PR: the 02.00 units still need retransmits on roughly half the cycles (the 02.10 unit needs none), but enough cycles complete that live data refreshes every 10–15 s. Happy to provide RF captures with the capture mode from this PR if useful.

Thanks a lot for this work - it took my two 02.00 units from unusable to fully operational.

Geoffn-Hub:
> **Tested on 3× HMS-2000-4T (North America / 915 MHz) - confirms the fix, with a hardware-revision data point**

This is interesting - the MIT-5000-8T that I have been using is Hardware Ver.H01.01.03, I have just had a newer one delivered so will see if I can test different hardware revisions on this model also.

Starfoxfs:
I have the same issue with an HMS-2000-4T (Firmware 01.01.17 and Hardware Version H00.04.10)

No stable connection only Fragments (Last Missing, Middle Missing in RX Debug)

Serial 1164A....

Board ESP32-S3 16MB with CMT2300A

Can someone compile this or make a a Alpha Version binary for OpenDTU ?

:Update !!: https://github.com/tbnobody/OpenDTU/pull/2962 works !!, i have compiled it with Visual Studio C.

I hope this will be implented in next Firmware

schwatter:
@Geoffn-Hub

I'm also testing your capture-mode branch. I tweaked the timing. And it is a little bit better after that.

SystemConfigParaCommand.cpp --> setTimeout(10000);
CommandAbstract.h  --> #define MAX_RETRANSMIT_COUNT 100

After that i get more successful packages and i think full eventlog when i click on the inverter "show events"
in the live tab. Now 15. Before just 3 or so.

```

Beginn | Ende | ID | Meldung
-- | -- | -- | --
06:07:02 | 06:07:02 | 1 | Wechselrichter gestartet
06:08:21 | 06:08:22 | 207 | MPPT-A: Eingangsunterspannung
06:08:21 | 06:08:22 | 208 | MPPT-B: Eingangsunterspannung
06:08:21 | 06:08:22 | 216 | MPPT-C: Eingangsunterspannung
06:09:46 | 06:09:47 | 207 | MPPT-A: Eingangsunterspannung
06:09:46 | 06:09:47 | 208 | MPPT-B: Eingangsunterspannung
06:09:46 | 06:09:47 | 216 | MPPT-C: Eingangsunterspannung
06:11:11 | 06:11:12 | 221 | Abnormal wiring of grid neutral line
06:12:39 | 06:12:40 | 208 | MPPT-B: Eingangsunterspannung
06:12:39 | 06:12:40 | 216 | MPPT-C: Eingangsunterspannung
06:14:04 | 06:14:05 | 207 | MPPT-A: Eingangsunterspannung
06:16:53 | 06:16:54 | 221 | Abnormal wiring of grid neutral line
06:18:16 | 06:18:17 | 208 | MPPT-B: Eingangsunterspannung
06:18:16 | 06:18:17 | 216 | MPPT-C: Eingangsunterspannung
06:19:37 | 06:19:38 | 207 | MPPT-A: Eingangsunterspannung
```

schwatter:
And radio statistics
Screenshot_20260622-201019_Chrome

Geoffn-Hub:
> @Geoffn-Hub
> 
> I'm also testing your capture-mode branch. I tweaked the timing. And it is a little bit better after that.
> 


which inverter are you using?

schwatter:
Ah sorry, it is now a mixed issuetopic. I use also the MIT-5000.

Starfoxfs:
This Branch works fine for newer HMS-2000 also, before i had no RF connection at all.

Firmware 01.01.17 and Hardware Revision 00.04.10

Zurrmaxe:
> > **Tested on 3× HMS-2000-4T (North America / 915 MHz) - confirms the fix, with a hardware-revision data point**
> 
> This is interesting - the MIT-5000-8T that I have been using is Hardware Ver.H01.01.03, I have just had a newer one delivered so will see if I can test different hardware revisions on this model also.

I have also one MIT5000 with the 01.01.03 Hardware and have a lot of mistakes und lost packages. Sometimes the good packages are not more the 25 %.
The MIT-5000 is 3 m from the DTU, my first HMT2250 5 m and the the second HMT2250 32 m from the dtu.
I use an external 868MHZ Antenne connected to the DTU. 
In the last years, i build a lot of DTUs for OpenDtu and Ahoy projects.
But always with different ESP32 incl ESP32S3 and different CMT2300A -Models.

In the last weeks, i use only one brunch with VS Code with them and tested the results.

I found out, the i have always different results.
Some of these mixed DTUs worked with the Mit5000, others not..

So I think  where is also  a problem with the timing between ESP32 and the CMT2300A.

There are lots of CMT2300A from different companys on the market with different HF connections.

And what about the Software/Hardware in the CMT2300A Chips ?

These Chips are not build für hoymiles 
invertes. 
We use it as a simple, cheap way, to get the infos from the inverters.

Maybe we should think about an other typ of  HF-chip.

For the future modells of hoymiles  the best way should be  to use the RS485 from original DTUProS with our OpenDTu projekts.

We should use the Modbus TCP. to read out everything like http/get/past or json.

It give us all Information we need and we must not think about HF Hardware and build complete DTUs




Alf

Zurrmaxe:
I found this :

https://wasilukm.github.io/hoymiles_modbus/api
https://github.com/tbnobody/OpenDTU/pull/582/commits

Zurrmaxe:
@schwatter
Today I tested your modifikations in capture-mode branch :
SystemConfigParaCommand.cpp --> setTimeout(10000);
CommandAbstract.h --> #define MAX_RETRANSMIT_COUNT 100

First, I tested my two ESP32 DTUs at a great distance from the inverters. These DTUs have the same ESP32WROOM, but different CMT2300A models and achieve different results.

Both DTUs were connected to the same external magnetic base antenna.

dtuTEST


Next, I set up the DTU2 next to two inverted units. This resulted (DTU2a) in even better results.

I'm currently testing a DTU with an ESP32S3 chip. Let's see what happens

The Result:

dtuTEST2

The ESP32S3 DTU works better.

The test was always conducted under the same conditions: Same software, same antenna, same position, but different hardware.

schwatter:
At the moment, my successful radio reception rate is only 38% as well. Maybe I've also been messing with the source code a bit too much.

But I've got the DTUs under control now. The first one is running with 3 × HMS 1800 at 864.25 MHz, and the second DTU is on 866.50 MHz. That way they no longer interfere with each other.

Zurrmaxe:
> At the moment, my successful radio reception rate is only 38% as well. Maybe I've also been messing with the source code a bit too much.
> 
> But I've got the DTUs under control now. The first one is running with 3 × HMS 1800 at 864.25 MHz, and the second DTU is on 866.50 MHz. That way they no longer interfere with each other.


In general, the code already works very well. We just need to find the right values/settings for the timing.

This timing is very poor with the MIT5000 and needs to be adjusted.
Fehler

It should be reduced, as it only causes unnecessary traffic. ButI just don't know where.

Now i hve only theses errors 

[19:31:07.943] I (7854671) hoymiles: Fetch inverter: 138290780488  ( => HMT2250T)
[19:31:08.446] I (7854672) hoymiles: Request SystemConfigPara
[19:31:08.446] I (7854673) hoymiles: Queue size - NRF: 0 CMT: 8
[19:31:08.446] I (7855175) hoymiles: RX Period End
[19:31:12.942] E (7855176) hoymiles: ERROR in SystemConfigPara: Received fragment size: 16, min expected size: 38
[19:31:12.942] W (7855176) hoymiles: Packet handling error
.....
[19:31:12.942] I (7859673) hoymiles: Fetch inverter: 138291500326  (=>HMT2250T)
[19:31:13.525] I (7859674) hoymiles: Request SystemConfigPara
[19:31:13.525] I (7859675) hoymiles: Queue size - NRF: 0 CMT: 7
[19:31:13.525] I (7860253) hoymiles: RX Period End
[19:31:13.685] E (7860254) hoymiles: ERROR in SystemConfigPara: Received fragment size: 16, min expected size: 48
[19:31:13.685] W (7860254) hoymiles: Packet handling error
....
[19:31:17.944] I (7864675) hoymiles: Fetch inverter: 1520a023df4e (=>MIT5000)
[19:31:19.937] I (7864676) hoymiles: Request SystemConfigPara
[19:31:19.938] I (7864676) hoymiles: Queue size - NRF: 0 CMT: 5
[19:31:19.938] I (7866666) hoymiles: RX Period End
[19:31:20.115] E (7866667) hoymiles: ERROR in SystemConfigPara: Received fragment size: 16, min expected size: 48
[19:31:20.116] W (7866667) hoymiles: Packet handling error

but other error says:

[19:33:03.007] I (7969735) hoymiles: Fetch inverter:  1520a023df4e (=>MIT5000)
[19:33:07.485] I (7969736) hoymiles: Request SystemConfigPara
[19:33:07.485] I (7969736) hoymiles: Queue size - NRF: 0 CMT: 8
[19:33:07.485] I (7974211) hoymiles: RX Period End
[19:33:07.658] E (7974212) hoymiles: ERROR in SystemConfigPara: Received fragment size: 16, min expected size: 38
[19:33:07.658] W (7974212) hoymiles: Packet handling error
Why are there different values? expected size: 38 ? then expected size: 48 ?

Alf

OlafWalther1969:
Das mit dem Schwachlichtverhaltenverhalten mit der Firmware V01.01.12-01.01.08 kann ich bestätigen.
Trotz 8 PV Module und 84V pro MPPT hat der Wechselrichter frühestens 1 Stunde nach den HMS und HMT Wechselrichtern mit der Stromproduktion begonnen.

Es gibt eine neue Firmware von Hoymiles V01.04.00-01.04.00.
Ich hatte gestern den Support Deutschland angerufen.
Die haben über den Hoymiles Server sofort die neue Firmware eingespielt.

Die neue Firmware wird nicht in der DTU Pro S angezeigt.
Dafür müsst ihr den Support anrufen.

Das neue Firmware merkt man sofort, die PV Leistung steigt um 5% und morgens beginnt der MIT-5000-8T fast schon früher wie die HMS und HMT Serie.

Zu der Startspannung kann ich aber noch nichts genaues sagen.


Grüße

Zurrmaxe:
@OlafWalther1969
Besten Dank für die Info über eine neue Hoymiles-Firmware. 
Halte uns bitte auf dem Laufenden, wie die Firmware ist und was besser geworden ist.
Hast du mal für mich den Link der Ahoy-Version ?  Will mal sehen, wie die es machen.

Ich betreibe meinen MIT5000 mit jeweils drei "älteren" 395W-Platten. 
Diese haben ein max. Spannung von 38 Volt und können somit im Regelbereich der MPPTs gut betrieben werden. 
Wichtig ist nur, dass die Summe der Leerlaufspannungen der drei Platten immer (auch im Frühling) unter den maximalen MPPT -Spannungen (140 Volt) bleiben. 
Somit konnte man ich mit einem MIT5000 meine zwei HMT2250T ersetzen.
lg Alf

OlafWalther1969:
@Zurrmaxe 
an den Eingängen von dem MIT-5000-8 sind 8 Trina angeschlossen, mit über 80V pro Eingang am Wechselrichter.
Mit der alten Firmware hat der MIT über 1 Stunde später mit der Produktion begonnen, als die HMT und HMS WR.

Das hat sich mit der neuen Firmware V01.04.00-01.04.00 vollkommen geändert.

smoe:
I have wired-up an 5000-MIT-8T today as a quick test if it is basically functional after having 3 30V modules at a single lane not making it blink yesterday. Today it first remained dead, but having those three modules at three different lanes then had it blink red, so it appears to fire up also without all lanes connected. I then added a fourth module, which I think to have sped up the blinking frequency. Connecting it to the grid then had the device blink green. 

I had a look at @Geoffn-Hub's two PRs and liked them.

Is there a way for me to help bringing this forward? It should be reasonably possible for me to rebase, compile and test these PRs, but there is only little incentive to just do it for myself.

smoe:
@Geoffn-Hub I followed up on my offer and prepared three handoff branches, moving both PR histories from their old common base (`f7e36933`) onto current `tbnobody/master` (`b6f0353b`, v26.3.30):

- #2963: original `31bf54e0` → rebased `663f657c`: [[rebase-geoffn-cmt-fifo-burst-drain](https://github.com/smoe/OpenDTU/tree/rebase-geoffn-cmt-fifo-burst-drain)](https://github.com/smoe/OpenDTU/tree/rebase-geoffn-cmt-fifo-burst-drain)
- #2962: original `227cc893` → rebased `8c7f1243`: [[rebase-geoffn-capture-mode](https://github.com/smoe/OpenDTU/tree/rebase-geoffn-capture-mode)](https://github.com/smoe/OpenDTU/tree/rebase-geoffn-capture-mode)
- Combined stack: `222b6af7`: [[supporting_5000_MIT_8T](https://github.com/smoe/OpenDTU/tree/supporting_5000_MIT_8T)](https://github.com/smoe/OpenDTU/tree/supporting_5000_MIT_8T)

The combined branch intentionally places #2963 first and then applies all 11 commits from #2962. #2962 already duplicates three parts of #2963—the FIFO drain, processing all buffered packets, and breaking when the buffer is full—but lacks #2963’s `PKT_OK`-only `available()` fix. I reconciled that overlap while preserving the complete #2962 series.

As a cross-check, the combined tree differs from the rebased #2962 tree only by that missing #2963 behavior.

The only conflict between #2962 and current master was the web API route registration. Current master’s handler casts were retained for the new capture endpoints.

All three branches pass a complete PlatformIO `generic_esp32` firmware build. I have not hardware-tested the rebased branches yet, but I can now flash the combined branch on my MIT-5000-8T and report the results.

Credit where it is due: OpenAI Codex did the heavy lifting here—history inspection, both rebases, conflict and overlap resolution, and build verification—under my direction and review.

I have not knowingly changed either of your PR branches; these are review and handoff pointers.

smoe:
### CMT burst processing reads and removes different queue elements

Codex found a queue-ordering bug in the CMT burst-drain implementation.

When `_packetReceived` is set, `HoymilesRadio_CMT::loop()` drains the CMT hardware FIFO. For every received packet it constructs a complete `fragment_t`, including its payload, length, channel and RSSI:

```cpp
fragment_t f;
memset(f.fragment, 0xcc, MAX_RF_PAYLOAD_SIZE);
f.len = std::min<uint8_t>(
    _radio->getDynamicPayloadSize(),
    MAX_RF_PAYLOAD_SIZE);
f.channel = _radio->getChannel();
f.rssi = _radio->getRssiDBm();
_radio->read(f.fragment, f.len);
_rxBuffer.push(f);
```

The important insertion is:

```cpp
_rxBuffer.push(f);
```

`_rxBuffer` is a `[std::queue](https://en.cppreference.com/w/cpp/container/queue)<fragment_t>`. According to the [`std::queue`[ documentation](https://en.cppreference.com/w/cpp/container/queue)](https://en.cppreference.com/w/cpp/container/queue), it is FIFO: `push()` inserts at the end, `front()` accesses the first element, `back()` accesses the last element, and `pop()` removes the first element.

The processing loop instead accessed `back()` and subsequently called `pop()`:

```cpp
fragment_t f = _rxBuffer.back();
// process f
_rxBuffer.pop();
```

Consequently, with queued fragments `[A, B, C]`:

1. `back()` processes C, while `pop()` removes A.
2. `back()` processes C again, while `pop()` removes B.
3. C is processed for a third time and then removed.

A and B are never processed. This directly defeats the purpose of draining multi-packet bursts and can prevent multi-fragment responses from being reassembled.

The minimal correction is to access the same oldest element that `pop()` removes:

```diff
-        fragment_t f = _rxBuffer.back();
+        fragment_t f = _rxBuffer.front();
```

This is about to be committed on `supporting_5000_MIT_8T` as:

```text
41e4c2dd fix(cmt): process queued RX fragments in FIFO order
```

The complete `generic_esp32` firmware build passes with this change.

smoe:
### SystemConfig buffer capacity became an unintended response requirement

Codex found a regression caused by using one size for two different purposes in `SystemConfigParaParser`.

Before the MIT changes, the SystemConfig buffer was 16 bytes and `getExpectedByteCount()` returned that size. PR #2962 correctly enlarged the buffer to 48 bytes so it could hold the longer MIT response:

```diff
-#define SYSTEM_CONFIG_PARA_SIZE 16
+#define SYSTEM_CONFIG_PARA_SIZE 48
```

It also introduced a configurable expected byte count, but initialized it from that enlarged buffer size:

```cpp
uint8_t _expectedByteCount = SYSTEM_CONFIG_PARA_SIZE;
```

The command does not compare this value for equality. It treats it as a lower bound:

```cpp
const uint8_t fragmentsSize =
    getTotalFragmentSize(fragment, max_fragment_id);

if (fragmentsSize < expectedSize) {
    return false;
}
```

Consequently, every inverter other than MIT inherited a minimum response requirement of 48 bytes. Valid legacy HM/HMS/HMT responses satisfying the previous 16-byte minimum could therefore be rejected.

These are separate concepts:

- **Buffer capacity:** the maximum amount of response data that can safely be stored.
- **Minimum response size:** the smallest reassembled response considered complete enough to parse.

The patch names and configures them separately:

```diff
-#define SYSTEM_CONFIG_PARA_SIZE 48
+#define SYSTEM_CONFIG_PARA_BUFFER_SIZE 48
+#define SYSTEM_CONFIG_PARA_DEFAULT_MIN_RESPONSE_SIZE 16
```

```diff
-uint8_t _expectedByteCount = SYSTEM_CONFIG_PARA_SIZE;
+uint8_t _minimumResponseSize =
+    SYSTEM_CONFIG_PARA_DEFAULT_MIN_RESPONSE_SIZE;
```

The API and validation code are renamed to describe the actual lower-bound semantics:

```diff
-uint8_t getExpectedByteCount() const;
-void setExpectedByteCount(uint8_t count);
+uint8_t getMinimumResponseSize() const;
+void setMinimumResponseSize(uint8_t size);
```

MIT retains its model-specific requirement:

```cpp
SystemConfigPara()->setMinimumResponseSize(38);
```

The resulting behavior is:

| Inverter | Buffer capacity | Minimum accepted response |
|---|---:|---:|
| HM/HMS/HMT | 48 bytes | 16 bytes |
| MIT-5000-8T | 48 bytes | 38 bytes |

Thus, the larger MIT response still fits safely, while existing inverter families retain the acceptance threshold used before the buffer expansion.

The fix is committed on `supporting_5000_MIT_8T` as:

```text
a813cc14 fix(hoymiles): separate SystemConfig buffer and response sizes
```

The complete `generic_esp32` firmware build passes with this change.

tbnobody:
@smoe thank you for the hint regarding the rx buffer. Havn't testet yet but it also affects the the nrf Part and could make things better! Would you mind to genervte a seperate PR just to fix this issue in cmt and nrf code?  (I can also create a fix and mention your Name. As you prefere)

smoe:
> @smoe thank you for the hint regarding the rx buffer. Havn't testet yet but it also affects the the nrf Part and could make things better! Would you mind to genervte a seperate PR just to fix this issue in cmt and nrf code?  (I can also create a fix and mention your Name. As you prefere)

@tbnobody, great to hear that this was putatively helpful beyond the MIT work. Please proceed directly. My literally only own contribution was to ask Codex "Is there anything that raises you attention?" and to then review what Codex came up with. Not sure if I should be credited for that. But I have a Codex-driven OpenClaw instance with a seperate GitHub account @smoe-bot, who I guess would be much appreciating to be mentioned.

smoe:
This one in my reading may be the major blocker for merging this MIT work.

### MIT recovery settings unintentionally affect every inverter

The MIT workaround changes two recovery settings for all inverter models:

1. The global fragment retransmission limit was increased from 5 to 20.
2. The non-MIT `RequestFrameCommand` response timeout was increased from the upstream 100 ms to 250 ms.

`RequestFrameCommand` is used whenever OpenDTU requests a missing response fragment. Consequently, a partial response from an existing HM/HMS/HMT inverter could occupy the shared radio for approximately:

```text
upstream:       5 × 100 ms = 0.5 s
current branch: 20 × 250 ms = 5.0 s
```

This can delay every other inverter and command in the queue by roughly ten times as long.

The patch introduces separate, centrally defined retransmission limits:

```cpp
#define MAX_DEFAULT_RETRANSMIT_COUNT 5
#define MAX_MIT_RETRANSMIT_COUNT 20
```

It selects the MIT limit only for serial prefix `0x1520` and restores the established 100 ms missing-fragment timeout for every other inverter:

```text
                    Maximum requests    Response timeout
HM/HMS/HMT          5                   100 ms
MIT-5000-8T         20                  2000 ms
```

The separate `MAX_RESEND_COUNT` setting, used when no fragments arrive at all, remains unchanged at 4.

```diff
diff --git a/lib/Hoymiles/src/commands/CommandAbstract.h b/lib/Hoymiles/src/commands/CommandAbstract.h
@@
 #define RF_LEN 32
 #define MAX_RESEND_COUNT 4 // Used if all packages are missing
-#define MAX_RETRANSMIT_COUNT 20 // MIT-5000-8T: 2 frags/burst, ~3-4 retransmits per remaining frag, need margin
+#define MAX_DEFAULT_RETRANSMIT_COUNT 5 // Used to send the retransmit package
+#define MAX_MIT_RETRANSMIT_COUNT 20 // Used to send the retransmit package to MIT inverters

diff --git a/lib/Hoymiles/src/commands/CommandAbstract.cpp b/lib/Hoymiles/src/commands/CommandAbstract.cpp
@@
 uint8_t CommandAbstract::getMaxRetransmitCount() const
 {
-    return MAX_RETRANSMIT_COUNT;
+    // Keep the experimental MIT recovery allowance local to that model.
+    constexpr uint16_t mitSerialPrefix = 0x1520;
+    const uint16_t serialPrefix = (_inv->serial() >> 32) & 0xffff;
+    if (serialPrefix == mitSerialPrefix) {
+        return MAX_MIT_RETRANSMIT_COUNT;
+    }
+
+    return MAX_DEFAULT_RETRANSMIT_COUNT;
 }

diff --git a/lib/Hoymiles/src/commands/SingleDataCommand.cpp b/lib/Hoymiles/src/commands/SingleDataCommand.cpp
@@
-    // MIT-5000-8T responds to retransmit requests at ~835ms intervals
+    // Keep the experimental MIT recovery timeout local to that model.
     const uint16_t prefix = (inv->serial() >> 32) & 0xffff;
     if (prefix == 0x1520) {
         setTimeout(2000);
     } else {
-        setTimeout(250);
+        setTimeout(100);
     }
 }
```

This isolation fix does not establish that 20 requests or 2000 ms are the correct MIT values. Those values are preserved to avoid changing MIT behaviour in the same patch, but they should be measured again after correcting the receive-queue bug.

The generic ESP32 firmware build succeeds with this patch.

Commit: `2c8b9b00 fix(hoymiles): scope recovery settings to MIT`

smoe:
Here is now a final "Priority 1" observation by Codex. This is not perfectly fixable until we truly know how the channel-changes work. Because of the queue-bug above I also did not want to induce the scheme from what was reported before. Codex asked me for an SDR. Not today.  :-)

@upstream, I prepared these patches as https://github.com/tbnobody/OpenDTU/pull/3161 for your convenience.

### Capture mode can make OpenDTU send a command on the wrong frequency

The CMT2300A is the radio chip that OpenDTU uses to communicate with HMS, HMT and MIT inverters.

A radio frequency is the frequency on which the radio sends or receives a message. The source code also calls each available frequency a “channel”.

This branch adds an optional function called capture mode. When capture mode is switched on, OpenDTU regularly changes the CMT2300A frequency and listens for messages from nearby inverters. It stays on each frequency for 50 milliseconds before trying the next one.

For normal communication, OpenDTU uses the value shown as **CMT2300A Frequency** in the DTU settings. In the source code, this value is stored in `_inverterTargetFrequency`. I will call it the communication frequency below.

The following sequence was possible:

1. Capture mode selected a frequency and listened for inverter messages there.
2. OpenDTU then found a normal command waiting to be sent.
3. The old send code did not first return the radio to the communication frequency.
4. OpenDTU therefore sent the command on the frequency that capture mode had selected.
5. After sending the command, OpenDTU returned to the communication frequency and waited for the inverter’s answer.
6. The inverter had probably not heard the command because it was listening on the communication frequency. It could therefore not answer.

The complete command was sent. It was not stopped while it was being sent. The problem was that it could be sent on the wrong frequency.

A later attempt could succeed because the radio had returned to the communication frequency by then. This could make the problem appear as an occasional lost command or a slow answer.

This problem is not limited to MIT inverters. HMS and HMT inverters use the same CMT2300A send code. Capture mode could therefore also disturb communication with these existing inverter models. The problem can only occur while capture mode is switched on.

### Capture mode can make OpenDTU send a command on the wrong frequency

The small correction is to select the correct frequency immediately before sending a command:

```diff
 if (cmd.getDataPayload()[0] == 0x56) {
     cmtSwitchDtuFreq(getInvBootFrequency());
+} else {
+    cmtSwitchDtuFreq(_inverterTargetFrequency);
 }
```

OpenDTU already treats one command differently.
Command `0x56` tells an inverter to change its communication frequency.
OpenDTU already sends command 0x56 on the start frequency stored for the selected geographic region. It then returns to the communication frequency selected in the DTU settings. This established OpenDTU behaviour predates the MIT changes and is left unchanged by this patch.

All other commands are now sent on the **CMT2300A Frequency** selected in the DTU settings.

Capture mode can still change frequencies while OpenDTU is only listening. It can also still miss messages because it listens on only one frequency at a time. This patch does not try to solve those limitations. It only prevents the frequency selected by capture mode from being used accidentally for a normal command.

smoe:
Upon request by @ms49434, key attributes have been abstracted away into the InverterAbstract class. This simplified the code and ensures more of a consistency - see https://github.com/tbnobody/OpenDTU/pull/3161/changes/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43 .

Zurrmaxe:
> Upon request by @ms49434, key attributes have been abstracted away into the InverterAbstract class. This simplified the code and ensures more of a consistency - see [5c1df91](https://github.com/tbnobody/OpenDTU/commit/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43) .

Hi smoe
I try to compile these new files with VS PlatformIO, but  have this error:

lib/Hoymiles/src/commands/CommandAbstract.cpp:132:12: error: 'MAX_RESEND_COUNT' was not declared in this scope
     return MAX_RESEND_COUNT;
            ^~~~~~~~~~~~~~~~
lib/Hoymiles/src/commands/CommandAbstract.cpp:132:12: note: suggested alternative: 'MAX_RF_FRAGMENT_COUNT'
     return MAX_RESEND_COUNT;
            ^~~~~~~~~~~~~~~~
            MAX_RF_FRAGMENT_COUNT


Alf

ms49434:
> > Upon request by @ms49434, key attributes have been abstracted away into the InverterAbstract class. This simplified the code and ensures more of a consistency - see [5c1df91](https://github.com/tbnobody/OpenDTU/commit/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43) .
> 
> Hi smoe I try to compile these new files with VS PlatformIO, but have this error:
> 
> lib/Hoymiles/src/commands/CommandAbstract.cpp:132:12: error: 'MAX_RESEND_COUNT' was not declared in this scope return MAX_RESEND_COUNT; ^~~~~~~~~~~~~~~~ lib/Hoymiles/src/commands/CommandAbstract.cpp:132:12: note: suggested alternative: 'MAX_RF_FRAGMENT_COUNT' return MAX_RESEND_COUNT; ^~~~~~~~~~~~~~~~ MAX_RF_FRAGMENT_COUNT
> 
> Alf

Please doublecheck if MAX_RESEND_COUNT is defined in CommandAbstract.h

line 10: #define MAX_RESEND_COUNT 4 // Used if all packages are missing

also, apply PR 3161: https://github.com/tbnobody/OpenDTU/pull/3161, not this one

schwatter:
Ok, yesterday the DTU Pro S arrived. I installed it quickly in the afternoon. At first, I noticed it needed an update because the MIT5000 wouldn't come online. After the update, it started receiving inverter data.

I sent an email to [service@hoymiles.com] asking them to update my inverter to firmware version V01.04.00-01.04.00. They updated it this morning, and the whole process took about half an hour.

Now I'm really surprised: since 11:30 this morning until now (13:43), I've had **100% successful reception**.

I'm still using Geoffn-Hub's `capture-mode` branch with `#define MAX_RETRANSMIT_COUNT 80`, but that was just for tinkering. I'll switch back to the default value of 20 tomorrow.

MIT5000-OpenDTU

Zurrmaxe:
First thanks to @ms49434.  

Here my old results from last month:

> In general, the code already works very well. We just need to find the right values/settings for the timing.
> 
> This timing is very poor with the MIT5000 and needs to be adjusted. Fehler
> 

> Alf

I build the new the firmware this morning and have this results:
Mit5000 new

Unfortunately, my levels have worsened.

Alf

smoe:
> Unfortunately, my levels have worsened.

Just for the sake of curiosity and since I cannot try myself - can you influence your success rate by getting the OpenDTU closer to the inverter (or move thad giant chunk of lead out of the way or unsheath the OpenDTU from its aluminium hat)?

schwatter:
Did you both read my comment? I would say the initial firmware is worse or the OpenDTU part won't fit for the firmware. I upgraded from V01.01.12-01.01.08 to V01.04.00-01.04.00. And now 100℅. Before i get also 33℅/33℅/33℅.

Zurrmaxe:
Hi everyone 😄 

The solution to our problem is quite simple:   **Firmware V01.04.00-01.04.00 ❗ ❗** 

Today in the morning i get the new firmware on my MIT-5000.

Here the results:
heute

DTU = ESP32S3 with the last modified code [5c1df91](https://github.com/tbnobody/OpenDTU/commit/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43)

Alf

smoe:
I just wish we would not need the vendor's DTU at all.

schwatter:
@Zurrmaxe 
I told you :) Btw i tested right now again with #define MAX_RETRANSMIT_COUNT 20 again.
So it is too low. 

Empfang Erfolgreich | 13 | 65 %
Empfang Fehler: Nichts empfangen | 0 | 0 %
Empfang Fehler: Teilweise empfangen | 7 | 35 %
Empfang Fehler: Beschädigt empfangen | 0 | 0 %

So #define MAX_RETRANSMIT_COUNT 40 is ok.

Empfang Erfolgreich | 10 | 100 %
Empfang Fehler: Nichts empfangen | 0 | 0 %
Empfang Fehler: Teilweise empfangen | 0 | 0 %
Empfang Fehler: Beschädigt empfangen | 0 | 0 %

smoe:
> @Zurrmaxe I told you :) Btw i tested right now again with #define MAX_RETRANSMIT_COUNT 20 again. So it is too low.

Looks much like it, indeed. I do not have an exact idea what the basis of that counting of packages is. But I could imagine some scenario with those channel changes and queue problems in which a package was not counted as missed or partly missed. Ideally, the MAX_RETRANSMIT_COUNT should be .. what .. 0 ? So having the need for such a high number of again-transmitted-requests is an indication of something being wrong.

Please kindly once try that [5c1df91](https://github.com/tbnobody/OpenDTU/commit/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43) branch with those 20.

Zurrmaxe:
> I just wish we would not need the vendor's DTU at all.

We only need the vendors's DTU for the Firmware update. 😒 

I used this settings:
#define RF_LEN 32
#define MAX_RESEND_COUNT 4 // Used if all packages are missing
#define MAX_DEFAULT_RETRANSMIT_COUNT 5 // Used to send the retransmit package
#define MAX_MIT_RETRANSMIT_COUNT 20 

Alf

smoe:
> > I just wish we would not need the vendor's DTU at all.
> 
> We only need the vendors's DTU for the Firmware update. 😒

There is admittedly an immediate reflex to reverse-engineer a firmware for the Hoymiles devices. Just like we are doing it for various Wifi routers - I am just not ultimately confident that I want any of those inverters to run on OpenWRT .-) . Preferably though we watch out for Open Source projects like 
 * µVerter: https://hackaday.io/project/204966-verter-the-open-source-solar-micro-inverter
 * LibreVerter: https://github.com/imendizabal/LibreVerter
 * https://github.com/transistorgrab/OSSI

schwatter:
> > @Zurrmaxe I told you :) Btw i tested right now again with #define MAX_RETRANSMIT_COUNT 20 again. So it is too low.
> 
> Looks much like it, indeed. I do not have an exact idea what the basis of that counting of packages is. But I could imagine some scenario with those channel changes and queue problems in which a package was not counted as missed or partly missed. Ideally, the MAX_RETRANSMIT_COUNT should be .. what .. 0 ? So having the need for such a high number of again-transmitted-requests is an indication of something being wrong.
> 
> Please kindly once try that [5c1df91](https://github.com/tbnobody/OpenDTU/commit/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43) branch with those 20.

This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.
But i think it is the "supporting_5000_MIT_8T" branch? I will build and test.

edit:
I build it. But the same with #define MAX_MIT_RETRANSMIT_COUNT 20

Gesendete Anfragen | 16 |  
Empfang Erfolgreich | 10 | 63 %
Empfang Fehler: Nichts empfangen | 0 | 0 %
Empfang Fehler: Teilweise empfangen | 5 | 31 %
Empfang Fehler: Beschädigt empfangen | 0 | 0 %
Gesendete Fragment Wiederanforderungen |  211

So i will go higher again to 40

edit2:
Maybe a problem with the refreshrate? I set it to 13sec in OpenDTU webinterface @ settings. Could be to fast.

smoe:
> > > @Zurrmaxe I told you :) Btw i tested right now again with #define MAX_RETRANSMIT_COUNT 20 again. So it is too low.
> > 
> > 
> > Looks much like it, indeed. I do not have an exact idea what the basis of that counting of packages is. But I could imagine some scenario with those channel changes and queue problems in which a package was not counted as missed or partly missed. Ideally, the MAX_RETRANSMIT_COUNT should be .. what .. 0 ? So having the need for such a high number of again-transmitted-requests is an indication of something being wrong.
> > Please kindly once try that [5c1df91](https://github.com/tbnobody/OpenDTU/commit/5c1df91b3ca0ddf0646755738d2f029e6c9e6b43) branch with those 20.
> 
> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository. But i think it is the "supporting_5000_MIT_8T" branch? I will build and test.

No, it is a fork of the repository, like the one you tried before that was from gueffn-hub, this one took gueff's work, rebased it on the current main and added some extras.

> edit: I build it. But the same with #define MAX_MIT_RETRANSMIT_COUNT 20

Thank you!

> Gesendete Anfragen | 16 |   Empfang Erfolgreich | 10 | 63 % Empfang Fehler: Nichts empfangen | 0 | 0 % Empfang Fehler: Teilweise empfangen | 5 | 31 % Empfang Fehler: Beschädigt empfangen | 0 | 0 % Gesendete Fragment Wiederanforderungen |  211
> 
> So i will go higher again to 40

So, for 16 packages the OpenDTU needs to ask 211+16=227 times and is successful only for 6 out of 10 packages is wanted to receive.

> edit2: Maybe a problem with the refreshrate? I set it to 13sec in OpenDTU webinterface @ settings. Could be to fast.

Yes, please do.

My temporary explanation is that something is wrong with the channel transition. I presume the OpenDTU asks just so often until that channel iteration by some random chance is working. Well, not having any immediate other ideas, so be it.

It would possibly be interesting to learn what channel was successful in receiving the message and the how-many-th request it was to which was successfully replied.

schwatter:
Even with 26sec i get:

Empfang Erfolgreich | 19 | 58 %
Empfang Fehler: Nichts empfangen | 0 | 0 %
Empfang Fehler: Teilweise empfangen | 13 | 39 %
Empfang Fehler: Beschädigt empfangen | 0 | 0 %
Gesendete Fragment Wiederanforderungen | 447

So again, i get back to #define MAX_MIT_RETRANSMIT_COUNT 40

Zurrmaxe:
Hi everyone, 😄 
It appears that the wireless statistics continue to run even when the inverter has stopped producing power.
This results in a lot of incorrect data.

The code should be modified to automatically perform a reset in the morning when the inverter starts up again.

Alf

Starfoxfs:
Sometimes it shows Inverter HMS2000 is offline because off to many bad transmits, but i can set it in Inverter Settings -> Advanced -> Accessibility threshold to a higher value then it not shows Offline.

HMS-2000 with Firmware 01.01.17

Here are my Results with https://github.com/tbnobody/OpenDTU/pull/2962 
on OpenDTUonBattery

Bildschirmfoto vom 2026-08-19 10-10-46



But before i doesnt have any connection to my HMS-2000 Inverter so this is an improvement.

smoe:
Revisiting (literally) the 5000-MIT installation of mine, I now first ever run that 5000-MIT branch on #3161 and was reasonably happy about it. It takes a while for the 5000-MIT to show up - but it shows. Codex and I have come up with some other bits, in particular with https://github.com/tbnobody/OpenDTU/pull/3188 that shows the three AC phases separately.

majk999:
> Czasami wyświetla się komunikat, że falownik HMS2000 jest offline, ponieważ jest zbyt wiele złych transmisji, ale mogę ustawić wyższą wartość w Ustawieniach falownika -> Zaawansowane -> Próg dostępności, wtedy nie będzie wyświetlał się jako offline.
> 
> HMS-2000 z oprogramowaniem układowym 01.01.17
> 
> Oto moje wyniki z #2962 na OpenDTUonBattery
> 
> Bildschirmfoto vod 2026-08-19 10-10-46
> Ale wcześniej nie miałem połączenia z moim falownikiem HMS-2000, więc to była poprawa.



> Sometimes it shows Inverter HMS2000 is offline because off to many bad transmits, but i can set it in Inverter Settings -> Advanced -> Accessibility threshold to a higher value then it not shows Offline.
> 
> HMS-2000 with Firmware 01.01.17
> 
> Here are my Results with #2962 on OpenDTUonBattery
> 
> Bildschirmfoto vom 2026-08-19 10-10-46
> But before i doesnt have any connection to my HMS-2000 Inverter so this is an improvement.
Hi @Starfoxfs,
I saw your comment and screenshot showing that you managed to run OpenDTU-OnBattery with PR #2962 merged/applied, which fixes the connection issue with the HMS-2000.
Could you please share your compiled ⁠.bin⁠ firmware file / build (or let me know how you built it)? I am struggling with the exact same issue with my inverter and would really appreciate your help.
Thanks in advance!

Starfoxfs:
Sure here it is:

[opendtuonbattery-20260518-mod.zip](https://github.com/user-attachments/files/31267543/opendtuonbattery-20260518-mod.zip)

This is compiled for ESP32-S3 Chips, i dont know if it works on another Chips

Firmware Update at your own Risk ! If something is broken I take no responsibility

ruckerzerg:
Can we merge this please to fix https://github.com/tbnobody/OpenDTU/issues/3063?

Starfoxfs:
I can say https://github.com/tbnobody/OpenDTU/pull/2962

Its not 100% stable on my HMS-2000-4T sometimes it loses connection and went Offline.

Some Finetuning is needed.

34% successfull
33% partial 
32% bad

Sometimes i get more Bad Packets so the OpenDTU went Offline.

ortegafernando:
is it true that ahoydtu has already implemented this ?

smoe:
> > @smoe thank you for the hint regarding the rx buffer. Havn't testet yet but it also affects the the nrf Part and could make things better! Would you mind to genervte a seperate PR just to fix this issue in cmt and nrf code?  (I can also create a fix and mention your Name. As you prefere)
> 
> @tbnobody, great to hear that this was putatively helpful beyond the MIT work. Please proceed directly.

I have now prepared this PR https://github.com/tbnobody/OpenDTU/pull/3198 with the queue handling fix.

Starfoxfs:
> is it true that ahoydtu has already implemented this ?

https://github.com/lumapu/ahoy/releases

AhoyDTU added HMS-1000 and have a button for CMT Communicated Inverters to see what Frequency the where Before Channel Hopping.

Test it, i dont know if it works fine with HMS Inverters or MIT Inverters

In my case i need OpenDTUonBattery because i have a Solar Buffer Battery.

PR #3008, opened 2026-03-30
Fix: WebSocket endpoints missing setAuthType for digest authentication

pieterhollander:
Fixes #3007


tbnobody:
Thank you. Already merged it in my local branch. Will therefor be included in the next release.

PR #3120, opened 2026-06-25
Add missing colon

ms49434:
The value of the html tag arial-controls is wrong. The expression "'v-pills-' + inverter.serial" must be expanded.


tbnobody:
Thank you. Will include it in the next release.

PR #3127, opened 2026-07-04
fix(build): patch framework to eliminate compiler warnings

stefan123t:
## Summary

Patches `esp32-hal-uart.c` in the ESP32 Arduino framework to fix two compiler warnings when building with strict flags (`-Wall -Wextra -Wnull-dereference`).

## Fixes

1. **Format specifier mismatch** (line 330): `%ld` used for `uint32_t baudrate` → changed to `%u` (ESP32 `uint32_t` is `unsigned int`, not `unsigned long`)
2. **Null pointer dereference** (line 460): `uart->num` dereferenced after `uart = NULL` → moved `log_e()` before the NULL assignment

## Implementation

Created `pio-scripts/patch_framework.py` — a pre-build script that patches framework files outside the git repository using string replacement with a marker to avoid double-patching.

The existing `patch_apply.py` uses `git apply` which cannot reach framework files in `~/.platformio/packages/`.

## References

- https://github.com/espressif/arduino-esp32/issues/9634 (milestone 3.0.0-RC3)
- Check if this patch can be removed when the PlatformIO platform bundles Arduino ESP32 core >= 3.0.0-RC3

### DIFF
--- a/pio-scripts/patch_framework.py
+++ b/pio-scripts/patch_framework.py
@@ -1,0 +1,67 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Patch framework files that are outside the git repository.
+# Uses string replacement instead of git apply.
+#
+import os
+import hashlib
+
+Import("env")
+
+FRAMEWORK_DIR = env.PioPlatform().get_package_dir("framework-arduinoespressif32")
+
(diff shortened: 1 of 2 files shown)


stefan123t:
should be fixed upstream in Arduino ESP32 core >= 3.0.0-RC3

PR #3101, opened 2026-06-12
Fix: CMT radio loses channel, PA level and RX state on country mode change

fpoisson2:
setFrequencyBand() fully re-initializes the CMT2300A register banks and leaves the chip in sleep mode. Any code path calling setCountryMode() therefore silently reset the working channel and TX power to the register bank defaults and stopped reception until the next transmission.

In particular, applyDataTaskCb() applied the PA level before the country mode, so every DTU settings save permanently downgraded the CMT TX power to the register bank default.

- Restore channel, PA level and RX state after setFrequencyBand()
- Remember the configured PA level so it can be re-applied
- Apply frequency and PA level after country mode in applyDataTaskCb()
- Switch the channel from standby (GoStby, set channel, GoRx) in setInverterTargetFrequency()

PR #3161, opened 2026-07-28
Supporting 5000 MIT 8T - Rebase of #2962 on master and extras

smoe:
As described on #2962.


lib/Hoymiles/src/commands/SingleDataCommand.cpp
ms49434, line 1: The serial number prefix shouldn't be hardcoded here, it makes it difficult to maintain the code when a new serial number range gets added.

Instead, the timeout value should come from the inverter object.
i.e.
add getSingleDataCommandTimeout method to Inverter and do

const uint32_t timeout = inv->getSingleDataCommandTimeout();
setTimeout(timeout);

lib/Hoymiles/src/commands/RealTimeRunDataCommand.cpp
ms49434, line 1: The serial number prefix shouldn't be hardcoded here, it makes it difficult to maintain the code when a new serial number range gets added.

Instead, the timeout value should come from the inverter object.
i.e.
add getRealTimeRunDataCommandTimeout method to Inverter and do

const uint32_t timeout = inv->getRealTimeRunDataCommandTimeout();
setTimeout(timeout);

lib/Hoymiles/src/commands/CommandAbstract.cpp
ms49434, line 1: The serial number prefix shouldn't be hardcoded here, it makes it difficult to maintain the code when a new serial number range gets added.

Instead, the maximum retransmit count value should come from the inverter object.
i.e.
add getMaxRetransmitCount method to Inverter and do

const uint8_t maxRetransmitCount = inv->getMaxRetransmitCount();
return maxRetransmitCount;

lib/Hoymiles/src/commands/AlarmDataCommand.cpp
ms49434, line 1: The serial number prefix shouldn't be hardcoded here, it makes it difficult to maintain the code when a new serial number range gets added.

Instead, the timeout value should come from the inverter object.
i.e.
add getAlarmDataCommandTimeout method to Inverter and do

const uint32_t timeout = inv->getAlarmDataCommandTimeout();
setTimeout(timeout);

lib/Hoymiles/src/commands/SingleDataCommand.cpp
smoe: Abstraction is now available.

lib/Hoymiles/src/commands/RealTimeRunDataCommand.cpp
smoe: Abstraction is now available.

lib/Hoymiles/src/commands/CommandAbstract.cpp
smoe: Abstraction is now available.

lib/Hoymiles/src/commands/AlarmDataCommand.cpp
smoe: Abstraction is now available.

smoe:
InverterAbstract.h now got additional attributes to consistently provide parameters for HM[ST] | MIT.

lib/Hoymiles/src/inverters/InverterAbstract.cpp
```diff
@@ -48,6 +49,26 @@ const String& InverterAbstract::serialString() const
...
+}
+
+uint32_t InverterAbstract::getAlarmDataCommandTimeout() const
+{
+    return 750;
+}
+
+uint8_t InverterAbstract::getMaxRetransmitCount() const
+{
+    return MAX_DEFAULT_RETRANSMIT_COUNT;
```
ms49434, line 69:
```suggestion
    return 5;
```

lib/Hoymiles/src/inverters/MIT_8CH.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+uint32_t MIT_8CH::getAlarmDataCommandTimeout() const
+{
+    return 12000;
+}
+
+uint8_t MIT_8CH::getMaxRetransmitCount() const
+{
+    return MAX_MIT_RETRANSMIT_COUNT;
```
ms49434, line 118:
```suggestion
    return 20;
```

lib/Hoymiles/src/commands/CommandAbstract.h
```diff
@@ -8,7 +8,8 @@
 
 #define RF_LEN 32
 #define MAX_RESEND_COUNT 4 // Used if all packages are missing
-#define MAX_RETRANSMIT_COUNT 5 // Used to send the retransmit package
+#define MAX_DEFAULT_RETRANSMIT_COUNT 5 // Used to send the retransmit package
+#define MAX_MIT_RETRANSMIT_COUNT 20 // Used to send the retransmit package to MIT inverters
```
ms49434, line 12: no longer needed when values are substituted in InverterAbstract.cpp and MIT_8CH.cpp

lib/Hoymiles/src/inverters/MIT_8CH.cpp
```diff
@@ -0,0 +1,143 @@
...
+ *   MPPT4: UDC=69.3V IDC=1.11A PDC=77.1W YT=151.566kWh
+ *   AC: UAC_1N=231.7V F=50.00Hz PAC=225.4W Q=0VAR PF=1.000
+ *   INV: T=12.0°C
+ *
+ * Unresolved offsets:
+ *   68: always 0 in captures so far (reserved?)
+ *   70: value matches PAC — possibly apparent power (S in VA)
+ */
+#include "MIT_8CH.h"
+#include "commands/CommandAbstract.h"
```
ms49434, line 33: Can be removed when MAX_MIT_RETRANSMIT_COUNT is substitited.

lib/Hoymiles/src/inverters/InverterAbstract.cpp
```diff
@@ -3,6 +3,7 @@
  * Copyright (C) 2022-2026 Thomas Basler and others
  */
 #include "InverterAbstract.h"
+#include "commands/CommandAbstract.h"
```
ms49434, line 6: Can be removed when MAX_DEFAULT_RETRANSMIT_COUNT is substituted

lib/Hoymiles/src/commands/AlarmDataCommand.cpp
ms49434: build was successful, thank you!

lib/Hoymiles/src/commands/CommandAbstract.cpp
ms49434: build was successful, thank you!

lib/Hoymiles/src/commands/RealTimeRunDataCommand.cpp
ms49434: build was successful, thank you!

lib/Hoymiles/src/commands/SingleDataCommand.cpp
ms49434: build was successful, thank you!

smoe:
@ms49434, I kind of like the #defines as these have all parameters nicely together and it is consistent with how everything is organised in the master branch. Also, to get access to the parameters you need an instance of those classes while you can literally grep in the code for those #defines (so have them while nothing is running) and at any point of the program's execution without any overhead.

To have them or not to have them, that to decide is a matter of upstream. The whole purpose of this branch is to render the MIT work by @Geoffn-Hub easier for upstream to adopt. Any extra changes of mine (or yours :-) ) would be detrimental to that cause, I fear. That is - unless if you are upstream.

Since I do not have immediate access to a functional OpenDTU (only the ones my neighbour failed to solder properly) or an MIT-8T - do you already have a success story?

ms49434:
My OpenDTU is running perfectly with a stripped version of this PR (RF capture code removed). I wanted to ensure that the modifications are 100% backward compatible with my existing inverters, mainly HMS-2000-4T. I own a MIT-5000-8T as well but haven't fired it up yet. Now that the code is looking more mature, I'll give it a try in the next days.

smoe:
It would be lovely if you could confirm that the RF capture code is now no longer of any harm for older models.

ms49434:
> It would be lovely if you could confirm that the RF capture code is now no longer of any harm for older models.

The full PR is running stable as well with capture mode enabled. No packet loss and no performance issues.
I'm running a Fusion board which has an ESP32S3 chip.

rudolfp1979:
Tested PR #3161 on OpenDTU-OnBattery with an HMS-2000-4T, serial prefix 0x1164, using an ESP32-S3 and CMT2300A.

Current radio statistics:

* Sent requests: 62
* Successful receptions: 62 (100%)
* Nothing received: 0
* Partially received: 0
* Corrupted receptions: 0
* Fragment retransmission requests: 650
* RSSI: -22 dBm

Before applying the PR, polling could stall for much longer and partial receptions occurred. With the PR, requests now complete reliably within approximately six seconds. There are still many fragment retransmission requests, but every request currently completes successfully and the CMT queue remains small.

smoe:
Thank you for the report. The many requests are in line with what other installations have reported. I have no immediate suggestions and for the foreseeable future lacks some combination of opportunity, equipment and skills to chase this up myself.

rudolfp1979:
If you Need some Support Tell me so i can try to do my best.

smoe:
## Update after field testing

I have updated the `supporting_5000_MIT_8T` branch from `5c1df91b` to `b7b2a724` with five additional commits prepared with help from Codex.

### Validate fragment lengths before calculating checksums

Commit `bfb74977` rejects RF fragments whose reported length is zero or exceeds the 32-byte RF payload buffer. It also checks that the final fragment of a multi-fragment response contains the two bytes required for its CRC16 checksum.

Without these checks, subtracting the checksum length from a malformed short fragment could underflow and make the checksum code read outside the received data.

This is the integration-branch version of #3163.

### Avoid advancing beyond an empty command queue

Commit `f602dd35` handles an empty command queue before the code calculates `begin() + 1`.

For an empty `std::deque`, `begin()` already equals `end()`. Advancing one position further is undefined behaviour. The new check is performed while the queue mutex is held.

Behaviour for queues containing one or more commands remains unchanged.

This is the integration-branch version of #3164.

### Reject oversized CMT2300A packets instead of truncating them

Commit `fc5cda74` preserves the payload size originally reported by the CMT2300A.

A packet larger than the supported RF packet size is still removed from the radio FIFO, but it is discarded instead of being passed to the parser as an apparently valid truncated packet.

### Select the MIT retransmission channel before receiving

Commit `a148048f` selects the receive channel used by the MIT when a missing fragment is requested again.

An actual fragment request is identified by both its command byte and its exact packet length. The observed channel pattern is:

- fragments 1, 4 and 7: inverter base frequency minus 250 kHz
- fragments 2, 5 and 8: inverter base frequency
- fragments 3, 6 and 9: inverter base frequency plus 250 kHz

The channel is selected before receive mode is started. Logging therefore cannot delay the short response.

Invalid fragment numbers and channels outside the supported range are rejected. Before every ordinary transmission, the radio is returned to the inverter's base channel. A previous MIT receive offset therefore cannot affect the following MIT or HMS command.

### Accept the SystemConfig response returned by the tested MIT

Commit `b7b2a724` removes the MIT-specific requirement for at least 38 reassembled SystemConfig bytes.

The tested inverter repeatedly returned a complete, CRC-valid, one-fragment response containing 16 bytes. Requiring 38 bytes rejected that valid response every time. The normal 16-byte validity requirement is therefore retained.

## Field-test result

The latest continuous recording produced:

- 28 of 28 successful MIT real-time-data requests
- successful reception of retransmitted fragments from all three frequency classes
- 95 of 95 successful commands from the first existing HMS inverter
- 107 of 107 successful commands from the second existing HMS inverter

MIT communication still requires several fragment retransmissions and therefore occupies the shared CMT radio for considerably longer than HMS communication. However, no interruption of an active HMS transaction, starvation, or HMS terminal failure was observed.

Capture mode remained disabled during these tests. The read-only polling path is well tested; MIT power-limit and other control commands have not yet been field-tested.

A clean `generic_esp32` build of `b7b2a724` succeeds:

- RAM usage: 23.9%
- flash usage: 92.5%

schwatter:
Screenshot 2026-08-21 192347

I've been running this build for two days now. This time I didn't change anything (like #define MAX_RETRANSMIT_COUNT)— just built and flashed it as is.

I'm getting a 100% success rate now. By the way, do you have access to your neighbor's MIT-5000 now?

The 20% "no capture" rate occurs during wakeup and sleep phases when OpenDTU receives no signal. It's the exact same behavior in Geoffn-Hub's source. Once the sun comes up and statistics reset, success goes back to 100%.

A quick question about frequency hopping: I thought it hops across three frequencies, but I'm only seeing two in my log. Am I missing something?

```
[19:30:25.793] D (176161121) webapi: Websocket: [/livedata][21] disconnect
[19:30:30.677] D (176164666) hoymiles: TX RealTimeRunData 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 80 0B 00 6A 88 8B 35 00 00 00 00 00 00 00 00 AC A7 83
[19:30:30.686] D (176164856) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 03 00 03 B9 63 09 2D 03 3B 00 1E 00 FE 00 04 9A BB DC | -60 dBm
[19:30:30.796] D (176165007) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 86 01 36 00 03 00 00 00 00 02 7A 00 02 00 00 66 07 76 | -60 dBm
[19:30:36.935] D (176170708) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 81 DE
[19:30:39.043] D (176172738) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 81 DE
[19:30:39.043] D (176172822) hoymiles: RX 867.75 MHz --> 95 A0 23 E1 9A 80 12 56 76 01 00 01 03 18 00 21 01 09 00 04 81 63 0B 84 03 3A BD | -60 dBm
[19:30:39.395] D (176174768) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 82 DD
[19:30:39.503] D (176174920) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 02 00 18 00 CA 00 03 CC 64 09 6F 03 27 00 17 00 BD 4C | -60 dBm
[19:30:41.474] D (176176800) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 84 DB
[19:30:43.539] D (176178831) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 84 DB
[19:30:43.539] D (176178917) hoymiles: RX 867.75 MHz --> 95 A0 23 E1 9A 80 12 56 76 04 0B D2 08 E5 08 E3 08 EF 0F 62 0F 6C 0F 77 13 87 01 | -60 dBm
[19:30:45.537] D (176180862) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 85 DA
[19:30:45.537] D (176180949) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 05 00 00 03 27 00 00 00 B3 00 0B 00 0C 00 0B 03 DB 99 | -60 dBm
[19:30:55.310] D (176190670) hoymiles: TX RealTimeRunData 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 80 0B 00 6A 88 8B 4F 00 00 00 00 00 00 00 00 CE BC 80
[19:30:55.493] D (176190898) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 03 00 03 B9 63 09 2D 03 45 00 1F 01 09 00 04 9A BB 55 | -60 dBm
[19:30:55.654] D (176191026) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 86 01 36 00 03 00 00 00 00 02 7A 00 02 00 00 B6 DE 7F | -60 dBm
[19:31:01.434] D (176196713) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 81 DE
[19:31:03.722] D (176198744) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 81 DE
[19:31:03.722] D (176198830) hoymiles: RX 867.75 MHz --> 95 A0 23 E1 9A 80 12 56 76 01 00 01 03 22 00 22 01 15 00 04 81 64 0B 85 03 42 E6 | -60 dBm
[19:31:07.652] D (176200775) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 82 DD
[19:31:07.652] D (176200861) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 02 00 19 00 D0 00 03 CC 64 09 6F 03 31 00 18 00 C7 34 | -60 dBm
[19:31:07.674] D (176202805) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 84 DB
[19:31:09.506] D (176204835) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 84 DB
[19:31:09.506] D (176204921) hoymiles: RX 867.75 MHz --> 95 A0 23 E1 9A 80 12 56 76 04 0B D2 08 E8 08 E1 08 EE 0F 63 0F 6C 0F 76 13 8A 02 | -60 dBm
[19:31:11.541] D (176206865) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 85 DA
[19:31:11.542] D (176206952) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 05 00 00 03 49 00 00 00 7A 00 0B 00 0C 00 0A 03 E8 0C | -60 dBm
[19:31:22.590] D (176216675) hoymiles: TX RealTimeRunData 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 80 0B 00 6A 88 8B 69 00 00 00 00 00 00 00 00 AF 0E 75
[19:31:22.590] D (176216866) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 03 00 03 B9 63 09 2D 03 3B 00 1F 01 01 00 04 9A BB 23 | -60 dBm
[19:31:22.603] D (176217015) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 86 01 36 00 03 00 00 00 00 02 7A 00 02 00 00 01 C7 D1 | -60 dBm
[19:31:27.409] D (176222718) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 81 DE
[19:31:29.431] D (176224749) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 81 DE
[19:31:29.431] D (176224835) hoymiles: RX 867.75 MHz --> 95 A0 23 E1 9A 80 12 56 76 01 00 01 03 0D 00 21 01 07 00 04 81 64 0B 85 03 38 A2 | -60 dBm
[19:31:31.467] D (176226779) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 82 DD
[19:31:31.467] D (176226866) hoymiles: RX 868.00 MHz --> 95 A0 23 E1 9A 80 12 56 76 02 00 18 00 CA 00 03 CC 64 09 6F 03 31 00 17 00 C3 24 | -60 dBm
[19:31:33.628] D (176228810) hoymiles: TX RequestFrame 868.00 MHz --> 15 A0 XX XX XX 80 12 56 76 84 DB
```

smoe:
Hm. Cannot tell, really. Capture mode is off, but that does not explain the two frequencies. The trace only re-requests fragments belonging to two frequency classes. The repeated requests are the real remaining inefficiency—but lowering MAX_MIT_RETRANSMIT_COUNT would merely cause earlier failures.

From what I understood, the frequency printed after RX is the channel on which the CMT2300A was listening. It is not an independent measurement of the inverter's transmitted frequency, which may be slightly off and received nonetheless. That would require an SDR recording to follow up and I am a bit too distracted with too many other things to for me to pursue this any time soon, I am afraid.

My perception was like yours: It works. And it even works when other inverters are communicating next to it. My access to the devices is only irregular, I am afraid. And I still need to order the SDR. :-/

ryan5836600:
Data point from an HMS-700-2T-NA (serial prefix 0x1146, rev A02) running your PR ported onto
OpenDTU-OnBattery — the three-channel fragment behaviour you model for the MIT-5000-8T appears
on this HMS too.

Pulling fragment IDs out of the RF dumps, the received sets in each initial burst are strict
arithmetic progressions with step 3:

```
AlarmData          got #3, #6, #9, #12
GridOnProFilePara  got #1, #4, #7, #10
RealTimeRunData    got #1 (of 3)
```

Across 185 single-fragment re-requests the DTU never once asked for #3, #6, #9 or #12 — those
always arrive. Not a link problem: RSSI −53 dBm, `RX Fail: Receive Corrupt` 0,
`RX Fail: Receive Nothing` 0. Single-fragment re-requests are answered ~30% of the time.

So the modulo-3 structure behind `getMitHopOffsetForFragment()` looks general, but the *phase*
doesn't match the fixed `((id - 1) % 3) - 1` formula on my unit — the captured residue class
differs between AlarmData (#3,6,9,12) and GridOnProFilePara (#1,4,7,10) in the same session.
So it can't just be ungated for 0x1146; it would need the phase learned per response rather
than derived from the fragment ID.

Separately, `a813cc14` ("separate SystemConfig buffer and response sizes") fixed a real bug for
me on plain HMS. Before it I got, every poll:

```
ERROR in SystemConfigPara: Received fragment size: 16, min expected size: 48
```

because the received size was compared against the buffer size rather than a minimum response
size. Worth noting it isn't MIT-only.

One more, possibly worth a separate report: with only ~1 fragment in 3 captured,
`MAX_DEFAULT_RETRANSMIT_COUNT` (5) can't assemble an 11–12 fragment response, so AlarmData and
GridOnProFilePara never completed (0/6 over six poll cycles) and were re-queued every cycle
forever. Raising that budget let both complete on the first try.

elmr91:
I'm current testing your PR with 2 HMS-2000-4T (identical HW and software releases).
One of them is particularly picky about transmissions:
HMS-2000 1 (-61 dB) : ~100% success no retransmit, no receive nothing
HMS-2000 2 (-66 db) : 66% success, 33% receive nothing, ~0 retransmit

Without this additional PR https://github.com/tbnobody/OpenDTU/pull/3101, I had to manually set the frequency after every ESP reboot (frequency / PALevel not correctly set after a reboot) otherwise nothing is received on my setup.
It may be worth including this change for people wanting to test this PR.

PR #3163, opened 2026-07-28
fix(hoymiles): validate fragment lengths before CRC

smoe:
## Summary

- reject empty or oversized RF fragments before calculating their CRC8
- reject multi-data responses whose final fragment does not contain the required two-byte CRC16

## Root cause

The radio receive paths can report a dynamic payload length of zero for a malformed packet. `HoymilesRadio::checkFragmentCrc()` previously subtracted one from that unchecked unsigned length. The value passed to `crc8()` therefore became 255, and the subsequent checksum-byte lookup used an index of `-1`.

There was a similar assumption in `MultiDataCommand::handleResponse()`: it subtracted two from the final fragment length without first verifying that the fragment contained its two CRC16 bytes.

## Impact

Malformed packets are now rejected through the existing error paths. Processing of valid packets is unchanged. The shared CRC8 guard protects both NRF and CMT radio implementations.


tbnobody:

PR #3164, opened 2026-07-28
fix(hoymiles): guard empty command queue

smoe:
## Problem

OpenDTU stores pending radio commands in a
[`std::deque`](https://en.cppreference.com/w/cpp/container/deque).

“Deque” means “double-ended queue” and is usually pronounced “deck”. It keeps
elements in order and allows efficient insertion and removal at both ends.

OpenDTU uses it as a first-in, first-out queue:

- New commands are added at the back.
- Handled commands are removed from the front.

C++ uses *iterators* to describe positions in the deque:

- `begin()` points to the first element.
- `end()` points one position after the last element. It is a boundary marker,
  not an element.
- For an empty deque, `begin()` and `end()` are equal.

`removeDuplicatedEntries()` and `replaceEntries()` deliberately start their
work at `begin() + 1`. This skips the first command, which may already be in
progress, and examines only later commands.

This is valid when the deque contains at least one element:

- With one element, `begin() + 1` equals `end()`. The resulting search range is
  empty.
- With more elements, `begin() + 1` points to the second element.

It is not valid when the deque is empty. `begin()` already equals `end()`, so
adding one advances beyond the permitted one-past-the-end position.

C++ calls this *undefined behavior*. This does not produce one defined error:
the program may appear to work, hang, crash, or behave differently after an
unrelated change.

The empty case can occur when the first active-power command is submitted.
OpenDTU tries to remove an older command before it adds the new command to the
queue.

## Fix

Return immediately when the deque is empty, before calculating
`begin() + 1`.

An empty deque contains nothing to remove or replace, so returning is also the
intended result.

The check is made while the queue mutex is held. Another task therefore cannot
empty the deque between the check and the iterator calculation.

The same guard is added to both functions that calculate `begin() + 1`:

- `removeDuplicatedEntries()`
- `replaceEntries()`

## Scope

This patch only handles the empty-deque case. It does not change command
ordering, duplicate handling, locking, or the special treatment of the first
queued command.

For every non-empty deque, the existing behavior remains unchanged.

## Validation

- `pio run -e generic_esp32` succeeds.
- `git diff --check` reports no whitespace errors.
- The zero-, one-, and multiple-element cases were reviewed:
  - Zero elements return before advancing the iterator.
  - With one element, `begin() + 1` still equals `end()`.
  - With multiple elements, processing still starts at the second element.


tbnobody:
Thank you. will Add it to next release.

PR #3166, opened 2026-07-29
Add RF info command to read the inverter's RF module version

rweickelt:
Fetches hardware and firmware version of the
inverter's built-in RF module using command 0x06
and exposes them in the web UI, the
/api/devinfo/status endpoint and two new MQTT
topics.

The request is repeated once per poll cycle until
an answer arrives, so inverters that do not
support it keep being asked for now.

Verified against: HM-1500 (nRF51802)


znuh:
Suggestion: Display the RfHw and RfFw versions in hex instead of decimal (%02x.%02x.%02x.%02x). That's apparently how the RF info is displayed in the official Hoymiles smartphone app:  
rf_ver

I started making a list of all known RF Modules with their RfHw & RfFw versions here: https://github.com/lumapu/ahoy/wiki/2.4GHz-(HM)-&-868MHz-(HMS-HMT-MIT)-RF-Modules
Perhaps we can maintain and supplement the RF documentation there.

tbnobody:
Thank you. Will add this PR to the next release.

@znuh I just changed the format string to the value you suggested.

This is my output:
image


Does this look reasonable for your?

znuh:
Yes, looks good! 👍

PR #3187, opened 2026-08-17
added inverter serial prefix for HMS_4T 1421

xX-nichtlachen-Xx:


tbnobody:

PR #3188, opened 2026-08-17
feature(webapp): show three-phase AC measurements

### LINKED ISSUE #1564: [Request] Show all 3-phases for HMT inverters in live view

As written by @Saarblick: The 3 phases are not shown separately in the live view.
> ... ein dreiphasiger HMT2250. Daher die 400V Die 3 Phasen werden ja nicht richtig von der DTU ausgewiesen... (https://github.com/hoylabs/OpenDTU-OnBattery/issues/1535#issuecomment-2603186005)

Image

Show all phases individually.

_No response_

### Additional context

_No response_

smoe:
## Summary

OpenDTU already decodes the phase-to-neutral voltages and phase currents for L1, L2, and L3 on supported HMT three-phase inverters. However, these measurements are not included in the live-data response.

The web interface therefore shows only generic voltage and current values and labels the AC output card as "Phase 1".

This change:

- adds the existing L1-N, L2-N, and L3-N voltage values to the live-data response;
- adds the existing L1, L2, and L3 current values;
- displays these six measurements in the AC output card;
- labels that card "AC output" when the phase details are available;
- hides the generic voltage and current rows in that case, because those values are aliases or combined values rather than another phase.

Existing live-data fields are retained for API compatibility. Inverters without the detailed phase fields keep their current display.

Total AC power, frequency, power factor, and reactive power remain unchanged. This PR does not expose the line-to-line voltages or calculate power for each individual phase.

## Scope

This is only an additive live-data API and web-interface change. It does not change radio communication, packet decoding, polling, or inverter control.

## Verification

- Prettier check passed
- ESLint passed
- Vue TypeScript check passed
- Vite production build passed
- Clean `generic_esp32` firmware build passed
- The existing `HMT_4CH` and `HMT_6CH` definitions on current master provide all six displayed fields

Physical HMT hardware was not available for testing this standalone branch, it was however tested together with #3161.


smoe:
image

Shown for Hoymiles 5000-MIT.

webapp/src/components/InverterChannelInfo.vue
```diff
@@ -16,14 +16,21 @@
         </div>
 
         <div v-if="channelType == 'AC'" class="card-header text-bg-info">
-            {{ $t('inverterchannelinfo.Phase', { num: channelNumber + 1 }) }}
+            <template v-if="hasPhaseDetails">{{ $t('inverterchannelinfo.ACOutput') }}</template>
+            <template v-else>{{ $t('inverterchannelinfo.Phase', { num: channelNumber + 1 }) }}</template>
```
AndreasBoehm, line 20: IMHO we can always use `AC Output` as the card title because `channelNumber + 1` will always be 1 and we will always show `Phase 1`, no matter to which phase a single inverter is connected to.

```suggestion
            {{ $t('inverterchannelinfo.ACOutput') }}
```

smoe: I completely agree. It is just a second change beyond the effort to introduce the three phases.

Issue #1629, opened 2024-01-08
After restart from mqtt opendtu does not send data to mqtt, restart is needet


svalcken:
### What happened?

mqtt server was rebootet. After them, opendtu does not reconnect and send data.

### To Reproduce Bug

Restart mqtt server and than data will not sendto mqtt.

### Expected Behavior

automatical reconnect to mqtt and send of data to mqtt.

### Install Method

Pre-Compiled binary from GitHub

### What git-hash/version of OpenDTU?

v23.12.28

tbnobody:
What do you see on the console if you restart the broker? What broker are you using? Please provide a screenshot of  **Info**  --> **MQTT**

bombjack7000:
I have that on OpenDTU v23.12.24, too, after restarting Home Assistant. I am using the standard MQTT server in Home Assistant (core-mosquitto).
Everything back to normal after restart of OpenDTU.
![grafik](https://github.com/tbnobody/OpenDTU/assets/75941026/06c0d0ee-a117-4de8-b76d-a454ace8f364)

tbnobody:
What is the mqtt status (info --> mqtt) if you restart the broker and the client does not work proberly?

bombjack7000:
Whoa,
OpenDTU is connected, MQTT Explorer also shows the MQTT packets coming in.
The problem seems to be on the Home Assistant side!

I better report that there.

Thank you for your quick reply.
OpenDTU is a great project! 👍 👍 👍 

![grafik](https://github.com/tbnobody/OpenDTU/assets/75941026/6fb91d6c-2a9d-4404-84f3-f31d8aaa0ff6)

srett:
I just had the same issue. **Info -> MQTT** showed **Status: Enabled**, but **Connection Status: Disconnected**. **Info -> Console** only showed inverter communication, nothing about MQTT. Simply saving the MQTT settings again without changing anything made it connect to mosquitto again. So it appears connection retry is broken under some circumstances. I'm on 24.1.26.

PR #3217, opened 2026-09-13
fix: WriteGuard deadlock

SW-Niko:
This fix will solve one rare WriteGuard() deadlock bug.

The fix use conditions on wait() to avoid:

* spurious wakeups
* missed wakeups
* a potential deadlock
* data races

How do we run into a deadlock?
Check the communication between the WriteGuard() Constructor/Destructor and the ConfigurationClass::loop().

grafik

<br>
<br>

(1) Configuration loop will wait until counter == 0;
(2) The second command arrives during write() and can not continue because the mutex is hold by the first command
(3) Notifying the loop, but the loop can not continue because the mutex is still hold by the first command
(4) Now the first command release the mutex

Who will get the mutex next? The second threat or the loop?
The C++ standard does not guarantee a specific order.
In case of the Thread 2:

* the counter will be set to 1
* the loop will wait for counter == 0. Will never happen. Wait for ever. --> deadlock
* the second command will wait for notification. Will never happen. Wait for ever. --> deadlock

We just run into this deadlock if a new thread acquires the mutex during the old thread is processed and mutex is locked.

The new pattern works like a "ship lock" 😜

- New ships can enter as long as the upper door is open and the lower door is closed
- New ships can't enter as long as ships in the "ships lock" leave the "ships lock" through the lower door
- After the "ships lock" is empty and the lower door is closed waiting ships will be informed to enter

I assume the bug is very rare in OpenDTU but in OpenDTU-OnBattery it seems we solved a major issue in combination with MQTT commands. See [#2560](https://github.com/hoylabs/OpenDTU-OnBattery/issues/2560), [#1629](https://github.com/hoylabs/OpenDTU-OnBattery/issues/1629)


tbnobody:
Thank you. Will add it in the next release.

mathieucarbou:
@SW-Niko @tbnobody : I ran also into this config deadlock A LOT while working on #3227 (Zero-Export feature).

My use case what that I am trying to a config update from a MQTT subsriber:

```
auto guard = Configuration.getWriteGuard();
auto& config = guard.getConfig();
[...]
```

The code is executed from the async_tcp task.

As I understand they had the same issue for OpenDTU-On-Battery.

This config lock mechanism is quite cryptic to me and I find it surprising because for a project on a MCU it seems like way overkill ;-) A function running a modifier anonymous function would have done the job with way less code / less risks for deadlocks and no loop mechanism ?

```
Configuration.update([](CONFIG_T& cfg) {
   cfg.foo = bar;
});
```

Then inside update we have the lock / unlock.

The idea is to isolate mutations.

In the current OpenDTU code, that's the opposite: mutations are not isolated and that's the ugly responsibility of the caller to get the lock to access the config, and what's more, to know about the internals and limitations of this locking mechanism.

Locking should be transparent, not a user facing api.

What's implemented is IMO way overkill and complex and I would really like to understand which use case is driving this complex design.

mathieucarbou:
> What's implemented is IMO way overkill and complex and I would really like to understand which use case is driving this complex design.

Ah! I missed that  the current design prevents async_tcp writers to race against loop task readers (i.e. reading config from loop task while writing it from async_tcp task). Using an updater does not protect against that.

But on the other end, what would prevent async_tcp task readers to not race against other Config updates from potentially elsewhere ?

SW-Niko:
Hello @mathieucarbou,

I do not use MQTT and do not update the configuration very often or at the same time from different web clients. 
So far, I haven't noticed any problems on my system with the WriteGuard().
Anyway, I stumbled upon the deadlock problem more or less by chance, because I analyzed the pattern in order to reuse it for similar data processing.

>I missed that the current design prevents async_tcp writers to race against loop task readers.

Yes, I assume that was the original design idea. But by now, we are reading from and writing to various threads.

>But on the other end, what would prevent async_tcp task readers to not race against other Config updates from potentially elsewhere?

I share your opinion, the WriteGuard() cannot prevent this. 

Just my opinion... 
Either we need to redesign the way configuration data is read and written to make it thread-safe,
or 
we need to stop using the configuration data for runtime updates via MQTT or similar external controls.

By the way, I have implemented some additional functions for OpenDTU-OnBattery and using the second option to avoid race conditions:
* I only read configuration data within the loop.
* I never update or write configuration data.

To get the required information, I inserted an additional layer between the executing task and the configuration data.
This additional layer uses a simple priority system to determine whether the required information is obtained from the configuration or from another source.

mathieucarbou:
> Either we need to redesign the way configuration data is read and written to make it thread-safe,

I agree. Currently, the implementation is error-prone, can lead to dead-lock and requires the user to know fro mwhich task he must and must not access / write the config. 

The only optimisation I like is the use of this Config.loop() function which is syncing the config access for all the tasks in the loop. That's a good optimisation because using shared_locks each time we do a Configuration.get() (in a returned wrapper) is costly.

We recommend in ESPAsycn32 to put async_tcp task on core 1 (it is by default on core 0) because users usually access application states within async_tcp callbacks.

https://esp32async.github.io/ESPAsyncWebServer/configuration/#important-recommendations-for-build-options

I went across all Config usages yesterday and I found that if OpenDTU was able to put the async_tcp task on core 1, most of the sync conditions could (maybe ?) be removed and eventually only keep a mutating update function containing a lock, because all config writes and reads happens in functions  / parts of code that will not likely yield or be pre-empted by higher priority task (Priority-Based Preemption) or same priority task (time slicing).

> we need to stop using the configuration data for runtime updates via MQTT or similar external controls.

I went this route with my Zero-Export feature (#3227).

When data is incoming through MQTT, I record the data in the class, set an atomic boolean flag, and in the main loop, do an exchange. If I can do the exchange, I then process the data on the loop. That's the only safe way for now.

```
void ZeroExportClass::onMqttMessageEnabled(const espMqttClientTypes::MessageProperties& properties, const char* topic, const uint8_t* payload, const size_t len)
{
    ESP_LOGD(TAG, "Received Zero-Export enabled command on %s", topic);
    const std::string value(reinterpret_cast<const char*>(payload), len);
    const bool enable = (value == "true" || value == "on" || value == "1" || value == "ON");
    _mqttEnabledValue.store(enable);
    _mqttEnabledPending.store(true);
}
```

Then in the task loop, I know that the configuration lock is help, so I only mutate the config live, does not save it, and access it through a forced cast:

```
// rendezvous guard from the main-loop task, which would self-deadlock.
static CONFIG_T& mutableConfig()
{
    return const_cast<CONFIG_T&>(Configuration.get());
}
```

This is a pain IMO.

tbnobody:
Just out of fun, I asked the evil AI. Thats the result. I added just  the WebApi_ntp example. All other locations where the config is read has to be adjusted acordingly.

```diff
diff --git a/include/Configuration.h b/include/Configuration.h
index 01b6406c8..5dcea27fc 100644
--- a/include/Configuration.h
+++ b/include/Configuration.h
@@ -2,9 +2,8 @@
 #pragma once
 
 #include "PinMapping.h"
-#include <TaskSchedulerDeclarations.h>
-#include <condition_variable>
 #include <cstdint>
+#include <functional>
 #include <mutex>
 
 #define CONFIG_FILENAME "/config.json"
@@ -182,34 +181,43 @@ struct CONFIG_T {
 
 class ConfigurationClass {
 public:
-    void init(Scheduler& scheduler);
+    void init();
     bool read();
     bool write();
     void migrate();
-    CONFIG_T const& get();
 
-    class WriteGuard {
+    // Grants read-only access to the configuration while holding the config
+    // mutex for the guard's lifetime. Safe to use from any task, including
+    // the AsyncTCP task that runs the web API callbacks.
+    class ReadGuard {
     public:
-        WriteGuard();
-        CONFIG_T& getConfig();
-        ~WriteGuard();
+        CONFIG_T const& getConfig() const;
+        INVERTER_CONFIG_T const* getInverterConfig(const uint64_t serial) const;
 
     private:
-        std::unique_lock<std::mutex> _lock;
-    };
+        friend class ConfigurationClass;
+        ReadGuard();
 
-    WriteGuard getWriteGuard();
+        std::lock_guard<std::recursive_mutex> _lock;
+    };
 
-    INVERTER_CONFIG_T* getFreeInverterSlot();
-    INVERTER_CONFIG_T* getInverterConfig(const uint64_t serial);
-    void deleteInverterById(const uint8_t id);
+    ReadGuard getReadGuard();
 
-    int8_t getIndexForLogModule(const String& moduleName) const;
+    // Runs fn with exclusive write access to the configuration, holding the
+    // config mutex for the duration of the call. Safe to call from any task,
+    // including the AsyncTCP task that runs the web API callbacks.
+    //
+    // Usage:
+    //   Configuration.update([](CONFIG_T& cfg) {
+    //       cfg.foo = bar;
+    //   });
+    void update(std::function<void(CONFIG_T&)> const& fn);
 
-private:
-    void loop();
+    static INVERTER_CONFIG_T* getFreeInverterSlot(CONFIG_T& config);
+    static INVERTER_CONFIG_T* getInverterConfig(CONFIG_T& config, const uint64_t serial);
+    static void deleteInverterById(CONFIG_T& config, const uint8_t id);
 
-    Task _loopTask;
+    static int8_t getIndexForLogModule(CONFIG_T const& config, const String& moduleName);
 };
 
 extern ConfigurationClass Configuration;
diff --git a/src/Configuration.cpp b/src/Configuration.cpp
index 61230191a..d7685df5a 100644
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -16,22 +16,22 @@ static const char* TAG = "configuration";
 
 CONFIG_T config;
 
-static std::condition_variable sWriterCv;
-static std::mutex sWriterMutex;
-static unsigned sWriterCount = 0;
-static bool sWriterActive = false;
-
-void ConfigurationClass::init(Scheduler& scheduler)
+// Protects all access to the global config struct. Config is written from
+// the AsyncTCP task (web API callbacks) and read from the main task
+// (scheduler loop) as well as from AsyncTCP, so every access has to go
+// through ConfigurationClass::ReadGuard / update().
+// Recursive because several functions acquire their own ReadGuard while a
+// caller further up the call stack already holds one on the same thread;
+// a plain mutex would deadlock in that case.
+static std::recursive_mutex sConfigMutex;
+
+void ConfigurationClass::init()
 {
-    scheduler.addTask(_loopTask);
-    _loopTask.setCallback(std::bind(&ConfigurationClass::loop, this));
-    _loopTask.setIterations(TASK_FOREVER);
-    _loopTask.enable();
-
     memset(&config, 0x0, sizeof(config));
 }
 
-bool ConfigurationClass::write()
+// Serializes the config to flash. Assumes sConfigMutex is already held.
+static bool writeConfigLocked()
 {
     File f = LittleFS.open(CONFIG_FILENAME, "w");
     if (!f) {
@@ -182,7 +182,14 @@ bool ConfigurationClass::write()
     return true;
 }
 
-bool ConfigurationClass::read()
+bool ConfigurationClass::write()
+{
+    std::lock_guard<std::recursive_mutex> lock(sConfigMutex);
+    return writeConfigLocked();
+}
+
+// Reads the config from flash. Assumes sConfigMutex is already held.
+static bool readConfigLocked()
 {
     File f = LittleFS.open(CONFIG_FILENAME, "r", false);
     Utils::skipBom(f);
@@ -363,7 +370,7 @@ bool ConfigurationClass::read()
     if (config.Dtu.Serial == DTU_SERIAL) {
         const uint64_t dtuId = Utils::generateDtuSerial();
         config.Dtu.Serial = dtuId;
-        write();
+        writeConfigLocked();
         ESP_LOGI(TAG, "DTU serial check: Generated new serial based on ESP chip id: %0" PRIx32 "%08" PRIx32 "",
             static_cast<uint32_t>((dtuId >> 32) & 0xFFFFFFFF),
             static_cast<uint32_t>(dtuId & 0xFFFFFFFF));
@@ -374,8 +381,16 @@ bool ConfigurationClass::read()
     return true;
 }
 
+bool ConfigurationClass::read()
+{
+    std::lock_guard<std::recursive_mutex> lock(sConfigMutex);
+    return readConfigLocked();
+}
+
 void ConfigurationClass::migrate()
 {
+    std::lock_guard<std::recursive_mutex> lock(sConfigMutex);
+
     File f = LittleFS.open(CONFIG_FILENAME, "r", false);
     if (!f) {
         ESP_LOGE(TAG, "Failed to open file, cancel migration");
@@ -463,24 +478,25 @@ void ConfigurationClass::migrate()
     f.close();
 
     config.Cfg.Version = CONFIG_VERSION;
-    write();
-    read();
+    writeConfigLocked();
+    readConfigLocked();
 }
 
-CONFIG_T const& ConfigurationClass::get()
+INVERTER_CONFIG_T* ConfigurationClass::getFreeInverterSlot(CONFIG_T& config)
 {
-    return config;
-}
+    for (uint8_t i = 0; i < INV_MAX_COUNT; i++) {
+        if (config.Inverter[i].Serial == 0) {
+            return &config.Inverter[i];
+        }
+    }
 
-ConfigurationClass::WriteGuard ConfigurationClass::getWriteGuard()
-{
-    return WriteGuard();
+    return nullptr;
 }
 
-INVERTER_CONFIG_T* ConfigurationClass::getFreeInverterSlot()
+INVERTER_CONFIG_T* ConfigurationClass::getInverterConfig(CONFIG_T& config, const uint64_t serial)
 {
     for (uint8_t i = 0; i < INV_MAX_COUNT; i++) {
-        if (config.Inverter[i].Serial == 0) {
+        if (config.Inverter[i].Serial == serial) {
             return &config.Inverter[i];
         }
     }
@@ -488,7 +504,7 @@ INVERTER_CONFIG_T* ConfigurationClass::getFreeInverterSlot()
     return nullptr;
 }
 
-INVERTER_CONFIG_T* ConfigurationClass::getInverterConfig(const uint64_t serial)
+INVERTER_CONFIG_T const* ConfigurationClass::ReadGuard::getInverterConfig(const uint64_t serial) const
 {
     for (uint8_t i = 0; i < INV_MAX_COUNT; i++) {
         if (config.Inverter[i].Serial == serial) {
@@ -499,7 +515,7 @@ INVERTER_CONFIG_T* ConfigurationClass::getInverterConfig(const uint64_t serial)
     return nullptr;
 }
 
-void ConfigurationClass::deleteInverterById(const uint8_t id)
+void ConfigurationClass::deleteInverterById(CONFIG_T& config, const uint8_t id)
 {
     config.Inverter[id].Serial = 0ULL;
     strlcpy(config.Inverter[id].Name, "", sizeof(config.Inverter[id].Name));
@@ -521,7 +537,7 @@ void ConfigurationClass::deleteInverterById(const uint8_t id)
     }
 }
 
-int8_t ConfigurationClass::getIndexForLogModule(const String& moduleName) const
+int8_t ConfigurationClass::getIndexForLogModule(CONFIG_T const& config, const String& moduleName)
 {
     for (uint8_t i = 0; i < LOG_MODULE_COUNT; i++) {
         if (strcmp(config.Logging.Modules[i].Name, moduleName.c_str()) == 0) {
@@ -532,47 +548,26 @@ int8_t ConfigurationClass::getIndexForLogModule(const String& moduleName) const
     return -1;
 }
 
-void ConfigurationClass::loop()
+CONFIG_T const& ConfigurationClass::ReadGuard::getConfig() const
 {
-    std::unique_lock<std::mutex> lock(sWriterMutex);
-    if (sWriterCount == 0) {
-        return;
-    }
-
-    // close the write guard for new writers, inform all existing writers and wait for them to finish
-    sWriterActive = true;
-    sWriterCv.notify_all();
-    sWriterCv.wait(lock, [] { return sWriterCount == 0; });
-
-    // open the write guard for new writers and inform all waiting writers that they can continue
-    sWriterActive = false;
-    sWriterCv.notify_all();
+    return config;
 }
 
-CONFIG_T& ConfigurationClass::WriteGuard::getConfig()
+ConfigurationClass::ReadGuard::ReadGuard()
+    : _lock(sConfigMutex)
 {
-    return config;
 }
 
-ConfigurationClass::WriteGuard::WriteGuard()
-    : _lock(sWriterMutex)
+ConfigurationClass::ReadGuard ConfigurationClass::getReadGuard()
 {
-    // if writers are currently processed, we must wait for them to finish before we can add ourselves to the writer count
-    if (sWriterActive) {
-        sWriterCv.wait(_lock, [] { return !sWriterActive; });
-    }
-
-    // add ourselves to the writer count and wait to be processed
-    sWriterCount++;
-    sWriterCv.wait(_lock, [] { return sWriterActive; });
+    return ReadGuard();
 }
 
-ConfigurationClass::WriteGuard::~WriteGuard()
+void ConfigurationClass::update(std::function<void(CONFIG_T&)> const& fn)
 {
-    sWriterCount--;
-    if (sWriterCount == 0) {
-        sWriterCv.notify_all();
-    }
+    std::lock_guard<std::recursive_mutex> lock(sConfigMutex);
+    fn(config);
 }
 
 ConfigurationClass Configuration;
+
diff --git a/src/WebApi_ntp.cpp b/src/WebApi_ntp.cpp
index 73de5b231..fa3633549 100644
--- a/src/WebApi_ntp.cpp
+++ b/src/WebApi_ntp.cpp
@@ -32,7 +32,8 @@ void WebApiNtpClass::onNtpStatus(AsyncWebServerRequest* request)
 
     AsyncJsonResponse* response = new AsyncJsonResponse();
     auto& root = response->getRoot();
-    const CONFIG_T& config = Configuration.get();
+    auto configGuard = Configuration.getReadGuard();
+    const CONFIG_T& config = configGuard.getConfig();
 
     root["ntp_server"] = config.Ntp.Server;
     root["ntp_timezone"] = Mycila::NTP.getTimezoneInfo();
@@ -73,7 +74,8 @@ void WebApiNtpClass::onNtpAdminGet(AsyncWebServerRequest* request)
 
     AsyncJsonResponse* response = new AsyncJsonResponse();
     auto& root = response->getRoot();
-    const CONFIG_T& config = Configuration.get();
+    auto configGuard = Configuration.getReadGuard();
+    const CONFIG_T& config = configGuard.getConfig();
 
     root["ntp_server"] = config.Ntp.Server;
     root["ntp_timezone_descr"] = config.Ntp.TimezoneDescr;
@@ -125,16 +127,13 @@ void WebApiNtpClass::onNtpAdminPost(AsyncWebServerRequest* request)
         return;
     }
 
-    {
-        auto guard = Configuration.getWriteGuard();
-        auto& config = guard.getConfig();
-
+    Configuration.update([&](CONFIG_T& config) {
         strlcpy(config.Ntp.Server, root["ntp_server"].as<String>().c_str(), sizeof(config.Ntp.Server));
         strlcpy(config.Ntp.TimezoneDescr, root["ntp_timezone_descr"].as<String>().c_str(), sizeof(config.Ntp.TimezoneDescr));
         config.Ntp.Latitude = root["latitude"].as<double>();
         config.Ntp.Longitude = root["longitude"].as<double>();
         config.Ntp.SunsetType = root["sunsettype"].as<uint8_t>();
-    }
+    });
 
     WebApi.writeConfig(retMsg);

```

Till now I wasn't aware of `recursive_mutex`.

tbnobody:
I just read  the comments for #2360 again. It was the goal to not acquire a mutex at each read access of the config. Therefor the diff which I showed above will maybe work as expected with the downside of the mutex at every config access.

mathieucarbou:
> I just read the comments for #2360 again. It was the goal to not acquire a mutex at each read access of the config. Therefor the diff which I showed above will maybe work as expected with the downside of the mutex at every config access.

Yes, acquiring on read would be really resource consuming, that's true.

what you have implemented above is similar to what I did implemented in a local branch also some days ago, except I only did the updater part, not the ReadLock part, because **it adds a lot of overhead.** + **there are a lot of places in OpenDTU where the config is read in single line return methods.**

I had a first version with shared_lock wrapped into a class this operator overload to directly access the config.
shared_locks allow to implement the ReadWriteLock pattern where you can allow multiple readers at the same time but guarantee exclusive access for writes. But shared_locks are expensive. 

locks or recursive locks are exclusive and would prevent 2 concurrent reads. So not good also.

An updater isolates the mutations, but sadly, only prevents 2 mutations to happen at the same time. 
It does not prevent a mutation to happen concurrently with a read from another core (which is bad) - in what I implemented, because I did not want to add overhead (locks or shared_locks) on reads.

so this is worse than the current implementation in OpenDTU I think...

the current code in OpenDTU at lest is optimised and does not add overhead on reads (there are many).

To reduce the likelyhood of concurrent access, I think you could already switch the async_tcp task on core 1 to make sure all callbacks, whether from mqtt (I guess you use the async version of the mqtt lib with async_tcp ?) or from async_ws will be run from core 1 only, like the user code (arduino loop task on core 1).

Then once it is done you need to pay attention to task preemption on same core: time-slicing preemption and priority-based preemption.

These can be temporary prevented with: taskENTER_CRITICAL() and taskEXIT_CRITICAL() to disable the system scheduler and interrupts,

```
// 1. Declare a portMUX_TYPE variable (usually globally or statically)
static portMUX_TYPE myMutex = portMUX_INITIALIZER_UNLOCKED;

void my_function() {
    // 2. Enter the critical section
    taskENTER_CRITICAL(&myMutex);

    /* --- CRITICAL SECTION --- 
       Keep this code as short and fast as possible.
       Do NOT use vTaskDelay(), print logs, or block here. */
    
    // 3. Exit the critical section
    taskEXIT_CRITICAL(&myMutex);
}
```

So maybe the solution could be:

1. Prevent preemption when doing reads (Config.get()) => this would be enough because you know that for the same core, the part of the code reading will be the only one executing and it will be fully executed.

2. add an updater function => this one ensures no concurrent writes.

3. put async_tcp on core 1 so that all the user code accessing the config is on the same core.

Downside of portMUX_TYPE: the code must be micro-second fast...

**Another better idea** 

- only keep the updater
- add a read function that reads under lock **but only use it for scalar types**

```
// e.g. in Configuration.h — copy-out under the existing writer mutex
template <typename T>
T getSnapshot(std::function<T(CONFIG_T const&)> const& fn)
{
    std::lock_guard<std::recursive_mutex> lock(sConfigMutex); // same mux as update()
    return fn(config);
}
```

Then, config reads are used in 2 ways:

- When reading basic types (atomic on 32 bits) => Direct call to Configuration.ge()...
- When using scalars: call getSnapshot()

This would ensure that reads stay concurrent (except scalar ones) and no write and scalar reads occur at the same time. Maybe a shared lock could be used also.

mathieucarbou:
@tbnobody : I raised a draft PR that compiles to show you what I had in mind: https://github.com/tbnobody/OpenDTU/pull/3237

```c++
    // Runs fn with exclusive write access to the configuration, holding the
    // config mutex for the duration of the call. Safe to call from any task,
    // including the AsyncTCP task that runs the web API callbacks.
    //
    // The return type is deduced from the callable: a lambda returning a
    // value forwards it to the caller, a lambda returning void simply
    // mutates. Plain lambdas work (same deduction pattern as read(Fn)).
    //
    // Usage (mutation):
    //   Configuration.update([](CONFIG_T& cfg) {
    //       cfg.foo = bar;
    //   });
    //
    // Usage (mutation with result):
    //   auto* slot = Configuration.update([](CONFIG_T& cfg) {
    //       return getFreeInverterSlot(cfg);
    //   });
    template <typename Fn>
    auto update(Fn&& fn) -> decltype(fn(std::declval<CONFIG_T&>()))
    {
        std::unique_lock<std::shared_mutex> lock(sConfigMutex);
        return fn(config);
    }

    // Consistent read: runs fn on the config while holding the shared lock,
    // so the values fn reads can never be torn by a concurrent update().
    // Multiple readers can run at the same time; writes stay exclusive.
    // fn must only read (not mutate) the config and must not block; use it
    // when several related fields must be captured atomically. For single
    // scalar fields, the lock-free Configuration.get() is fine.
    //
    // Usage:
    //   auto pos = Configuration.read([](CONFIG_T const& cfg) {
    //       return std::make_pair(cfg.Ntp.Latitude, cfg.Ntp.Longitude);
    //   });
    template <typename Fn>
    auto read(Fn&& fn) const -> decltype(fn(std::declval<CONFIG_T const&>()))
    {
        std::shared_lock<std::shared_mutex> lock(sConfigMutex);
        return fn(config);
    }
```

1. goes through an update fn to update the config using a unique lock
2. goes through a read fn to read parts of the config using a shared lock to allow multiple concurrent reads
3. don't use any mechanism: just use bare Configuration.get() when want to read a primitive type (32 bits max)

This limits the shared_locks usages only at places where multiple reads occur.

For 3 to work fine, you would need to make sure that the code mutating and accessing the config is on the same core, so which means put the async_tcp task on core 1. Otherwise you need to sync ALL reads also because there is no guarantee that core 1 will pick the value set by core 0 if it is cached for example.

This is a draft PR. Not tested.

tbnobody:
For the next release I would stick at this (#3217) PR and would suggest to rework the configuration afterwards.

mathieucarbou:
> For the next release I would stick at this (#3217) PR and would suggest to rework the configuration afterwards.

Definitely agree! This PR fixes the current sate, until a better rework is done.

PR #3218, opened 2026-09-14
Fixed High Byte alarm ID issue

xX-nichtlachen-Xx:
Error 308 was shown as 52 in ui:
Fix: AlarmLogParser truncates Event IDs ≥ 256 to their low byte
AlarmLogParser::getLogEntry() already assembles the full 16-bit wcode correctly (and uses it correctly for the AM/PM flag bits), but then sets entry.MessageId from the raw low byte of the buffer directly instead of from wcode. Any inverter alarm code ≥ 256 gets silently truncated to code & 0xFF, landing on an unrelated/undocumented number and displaying as "Unbekannt"/"Unknown" even though the code is already in _alarmMessages.

Real example, captured live from an HMT-2250W over radio (AlarmData response, second log entry):

raw bytes:  01 34 00 02 6A 7E EB 74 04 C5 0A D1
wcode     = 0x0134 = 308  ("INV overcurrent or overvoltage protection failure")
Start     = 09:34:22
End       = 18:44:36
Before this fix: entry.MessageId = payload[1] = 0x34 = 52 → not found in the table → displayed as 09:34:22 18:44:36 52 Unbekannt.
After this fix: entry.MessageId = wcode & 0x0FFF = 308 → matches the existing table entry → displayed as 09:34:22 18:44:36 308 INV overcurrent or overvoltage protection failure.
The fix only widens the Event ID extraction from 8 to 12 bits (still masking off the documented AM/PM flag bits at 12/13), so every code < 256 is unaffected and decodes identically to before.

Counter checked with JTAG on DSP.


xX-nichtlachen-Xx:
Before:
grafik

After:
grafik

xX-nichtlachen-Xx:
Error Codes 301 - 314 in Cloud:
grafik

tbnobody:

PR #3225, opened 2026-09-18
feat(hoymiles): early RX exit for faster inverter control

mathieucarbou:
## Problem

A radio transaction is currently only evaluated once its per-command RX
timeout expires, even when the inverter has already answered completely.
This is most noticeable for device-control commands (power limit,
power on/off), which use a 2000ms timeout: an acknowledgement typically
arrives within ~100ms, but the radio stays busy for the remaining ~1900ms
before the next queued command can be sent.

This makes OpenDTU comparatively slow for use cases that repeatedly send
power limit updates (e.g. dynamic/zero-export power control from an external
regulator such as Home Assistant or OpenDTU-on-Battery), since each update
costs a full timeout instead of the actual round-trip time.

## Change

A transaction is now finalized as soon as all of its expected fragments have
been received, instead of waiting for the timeout in every case:

- The last fragment of a response carries the 0x80 flag (already tracked in
  `_rxFragmentMaxPacketId`). Once that flag has been seen and every fragment
  up to it is present, completion is mathematically certain.
- This is exposed as a new side-effect-free
  `InverterAbstract::isTransactionComplete()`. Unlike `verifyAllFragments()`,
  it does not touch the retransmit counters, so it can be checked on every
  loop iteration without affecting statistics or retry logic.
- The radio's own RX fragment buffer must be fully drained
  (`isRxBufferEmpty()`, implemented per radio) before finalizing, so a
  fragment belonging to the current response can never be misattributed to
  the next transaction.
- The existing RX timeout is kept unchanged as a fallback for incomplete or
  missing responses, so slow inverters, retransmissions and error handling
  behave exactly as before.

Single-frame device-control acknowledgements (power limit, power on/off)
qualify for early exit naturally, since their response is a single 0x81
frame. This reduces the effective cost of a power limit command from the
full 2000ms timeout down to the actual round-trip time (typically
~100-150ms).

`ChannelChangeCommand` is unaffected: it is fire-and-forget by design (no
response is ever expected), so `isTransactionComplete()` never returns true
for it and it keeps finalizing via its existing 10ms timeout.

**Both changes make OpenDTU suitable for fast zero-export regulation loops
(e.g. Home Assistant reading a grid meter and publishing limit setpoints via
MQTT): a limit command now completes in its actual round-trip time
(~100-150ms) instead of the full 2000ms timeout, and it is sent before any
pending read requests.**

**This set of changes makes OpenDTU as fast as Ahoy for power control (limits).**

This PR is based on  #3226 and includes its commit.

## Compatibility

- No change to the CMT2300A/NRF24 drivers or their RX state machines.
- No behavior change for incomplete, missing, or slow responses.
- No change to command ordering, retry counts, or statistics semantics.

## Impact

Power limit command: full 2000ms timeout → actual round-trip (~100-150ms).
Regulation cycle across 8 HMS inverters: ~16s → ~1s.

## Testing

Tested at home on my 8 HMS-1000. 

- **Before**: the control loop was taking about 16 seconds to complete and the queue was always full of both commands and read requests
- **After**: the control loop is fast: I can regulate 8 HMS in less than 2 seconds. The queue execution is faster so items are dequeues also faster. 

As a consequence:
- all the power control commands take less than 2 sec to execute
- all the readings are fast
- I can safely decrease the poll interval to 2 seconds to ensure faster reads and in any case, power control commands will be executed with a higher priority.

image

**Some logs showing the changes**

[opendtu-log-2026-09-18_16-53-27.txt](https://github.com/user-attachments/files/32386738/opendtu-log-2026-09-18_16-53-27.txt)


mathieucarbou:
@tbnobody @schlimmchen FYI. 

I know several people using OpenDTU for power limit control through Home Assistant.
Some have switched to Ahoy because of this issue. OpenDTU is really slow to regulate power when we need zero injection.

I've read the github history, why these timeouts were set high. No objection to that.
But if we receive an early ack, why still waiting.

I hope this PR will make in. That's a very good improvements in OpenDTU reactivity!

Note: a good thing of this PR is that MQTT validation could be dropped to 1 or 2 seconds in order for Home Assistant to get the feedback about the invert stats updates faster also.

mathieucarbou:
converted to draft - I saw this morning that all inverters were unreachable in OpenDTU. Checking what could cause that.

mathieucarbou:
I completed some testing thanks to PR https://github.com/tbnobody/OpenDTU/pull/3227.

## Regulation before this fix:

Capture d’écran 2026-09-19 à 15 43 59

- DTU poll interval : 5 sec
- Regulation each 30 sec

With 8 inverters this is not possible top improve more.

## After this fix:

Capture d’écran 2026-09-19 à 16 17 16

image

image

Queue is drained quickly and power control are sent within 2 sec.

mathieucarbou:
Converted to draft again. Here is the status:  works fine over 2 days on 3 users having HMS.

But I would like to test it under specific conditions and different rates with the Zero-Export PR which is a good use case to test it.

I will report back in a few days and move the PR to draft once I consider it stable.

For now, it is running fine without issue for the persons I know who have tested.

tbnobody:
I already merged #3166 which also introduced commands which don't end up with 0x80. Maybe we need a exception in this program flow as well.

mathieucarbou:
> I already merged #3166 which also introduced commands which don't end up with 0x80. Maybe we need a exception in this program flow as well.

FYI 3 people I know of are testing this change since 3 days so far no issue :-)

For the commands which don't end up with 0x80, if you know the end you can add, otherwise the default timeout in place will be used to determine the end.

tbnobody:
Yes, but early exit will not work for responses without '0x80' . Question is if this has any negative side effects when such responses are received. Without thinking deep into it, I would assume that the timeout will still be triggered. (Which is also Ok because the mentioned command is only executed once)

mathieucarbou:
> Yes, but early exit will not work for responses without '0x80' . Question is if this has any negative side effects when such responses are received. Without thinking deep into it, I would assume that the timeout will still be triggered. (Which is also Ok because the mentioned command is only executed once)

Yes that is true. 

But this PR does not modify the existing behaviour. It just allows to not wait until the timeout finishes when we know com is finished. Some inverters do not answer with 0x80 and that's why for them the code will have to wait until the timeout ends like now. There is no change for them.

So instead of slowing down everybody, then, at least, for inverters that answer correctly, an early exit is proposed. 

This is useless to continue to wait when we know the inverter has answered right ?

Especially for a Zero-Export feature, taking 2 sec per request to apply a limit is huge and inefficient: 10 inverters make up for 20 seconds to apply a new power limit, when they can all be applied within 2 seconds if we exit early **when we can**.

tbnobody:
Just read the code again and merged it. Will be included in next release.

SW-Niko:
Hello @mathieucarbou , @tbnobody ,
In general, I really like the idea of ​​speeding up the regulation.
However, I fear that by doing so, we are opening the door to problems associated with control loops.

Let me explain why.

Some commands, such as the "power limit command" have not yet been completed by the time we receive the response from the inverter. It takes about 1s, before the new limit is settled on the AC output of the inverter. Right now we can trust the timing of the "power limit command". Once the command is complete, we know we have the correct AC power.

And now, we transfer the time behavior of the control loop more or less to the external system and leave the correct timing to the external programmer. And if we issue commands too quickly, we encounter control problems such as overshoot, undershoot, and oscillation.  

Don't get me wrong: I am prepared to deal with the new timing characteristics and possess knowledge of control loops.
Furthermore, we would force forks of the repository to adopt the new timing behavior immediately.

Maybe we must mark this as a "major change" 🤔

mathieucarbou:
> Hello @mathieucarbou , @tbnobody , In general, I really like the idea of ​​speeding up the regulation. However, I fear that by doing so, we are opening the door to problems associated with control loops.
> 
> Let me explain why.
> 
> Some commands, such as the "power limit command" have not yet been completed by the time we receive the response from the inverter. It takes about 1s, before the new limit is settled on the AC output of the inverter. Right now we can trust the timing of the "power limit command". Once the command is complete, we know we have the correct AC power.
> 
> And now, we transfer the time behavior of the control loop more or less to the external system and leave the correct timing to the external programmer. And if we issue commands too quickly, we encounter control problems such as overshoot, undershoot, and oscillation.
> 
> Don't get me wrong: I am prepared to deal with the new timing characteristics and possess knowledge of control loops. Furthermore, we would force forks of the repository to adopt the new timing behavior immediately.
> 
> Maybe we must mark this as a "major change" 🤔

Keep as-is would work if we could sent all power limit commands concurrently: when the control loop runs, if it has to send 10 commands concurrently and wait 2 sec for a timeout, that's fine. That' 2 sec max.

But the reality is that commands are sent in sequence so by the time the last command will be sent, the other ones will be applied.

So we only need to account for 2 second in the control loop AFTER the last command was sent. An easy way is to prevent the regulation loop to be too short.

For example in my implementation I prevent the user from setting it lower than 5 seconds, which is enough to let room for the power change to be applied.

image

Also, do not forget that what is driving a control loop is mostly the measurements and their pace. So if you have a control loop shorter than the interval at which the measurements are coming, you can apply a sort of source-adaptive correction based on the intervals.

So I think that's totally fine.

What is not acceptable is the current behavior: commands are sent in sequence and waiting for a timeout of 2 sec for a command to be applied, when having 10 inverters make up for 20 seconds, so 22 seconds in total to have a power control applied when all can be done in 2 seconds! That was driving me crazy lol! No wonder why I know some people switched to Ahoy who does not wait for this timeout value but early exit instead.

PR #3226, opened 2026-09-19
feat(hoymiles): track commands through the radio queue with unique IDs

mathieucarbou:
Every command instance now gets a unique id at creation time. All radio queue lifecycle log lines carry the id, command name and inverter serial, making it possible to follow a single command through enqueue -> TX -> RX period end -> success/failure:

  Enqueue [42] ActivePowerControl (HMS, 01) (1410a01e2cdd) (queue: 3)
  TX Start [42] ...
  RX Period End [42] ...
  Success [42] ...

Additionally the queue contents are dumped at each poll cycle so that pending entries (including their priority order) can be inspected.

This complements the priority-aware queue: since the emission order is no longer FIFO, log correlation by command identity is required to debug slow or failed transactions.

Log example:

[opendtu-log-2026-09-18_16-53-27.txt](https://github.com/user-attachments/files/32386738/opendtu-log-2026-09-18_16-53-27.txt)


mathieucarbou:
I have converter this PR to draft.

It is super useful to debug what is going on within the queue because the current logs are basically too simple to be helpful at all and do not help understand easily which message was executed / dequeued.

But it adds resource usage on the com stack for logging (which is a heavy task) and also does not help heap fragmentation.

I plan on keeping it running for a while over a few days, but I guess once my zero export feature will be completed I won't need it anymore.

PR #3227, opened 2026-09-19
feat: OpenDTU Zero-Export (supports Shelly LNM or MQTT)

mathieucarbou:
# feat: OpenDTU Zero-Export (supports Shelly LNM or MQTT)

## Summary

This PR adds a native Zero-Export controller to OpenDTU.

It regulates Hoymiles inverter power limits directly from the grid power measurement, removing the need for an external Home Assistant automation loop.

The regulation follows the same feedback model previously used in Home Assistant:

```text
missedPower = gridPower - setpoint
newLimit = previousRequestedLimit + missedPower
```

The correction is distributed proportionally across the available inverters and constrained by:

- Each inverter's nominal power
- The configured minimum production power
- Inverter reachability
- The command-enabled state
- The configured regulation interval

The regulation does not depend on inverter production telemetry because those values may be delayed or stale.

## Grid Sources

Grid sources are implemented through the `ZeroExportGridSource` interface. This keeps the regulation algorithm independent from the transport and makes additional sources easy to add later.

Currently supported sources:

- **Shelly Local Network Messaging**
- **MQTT**

Future sources such as Victron Modbus TCP or Fronius Modbus TCP can be added without changing the regulation logic.

### Shelly LNM

Shelly Local Network Messaging is Shelly's local UDP multicast protocol. Shelly devices publish measurements such as active power, voltage, current, power factor, and frequency without requiring HTTP polling.

OpenDTU:

- Joins the configured multicast group using `WiFiUDP`
- Processes packets from the scheduler without blocking
- Parses the Shelly LNM binary header and JSON payload
- Extracts `act_power` from Shelly EM and Shelly 3EM status messages
- Marks the source unavailable when no fresh packet is received

Default configuration:

- Multicast group: `239.255.55.55`
- UDP port: `5555`
- Freshness timeout: 10 seconds

### MQTT

The MQTT source uses OpenDTU's existing MQTT client and subscribes to a configurable grid-power topic.

Supported payloads:

- Plain numeric values, for example `-123.4`
- JSON objects containing `act_power`
- JSON objects containing `total_act_power`

This supports simple MQTT power sensors and Shelly EM / Shelly 3EM-style JSON payloads.

Retained MQTT messages are ignored because they may contain an old grid value that must not drive regulation.

The MQTT source uses a 60-second freshness timeout, allowing slower MQTT publishers than Shelly LNM.

## Inverter Power Limits

Zero-Export only sends non-persistent limits:

```cpp
PowerLimitControlType::AbsolutNonPersistent
```

The persistent inverter limit is never modified.

The controller tracks the last absolute limit requested by Zero-Export and uses that value as the starting point for the next regulation step.

### Startup and Reboot

When Zero-Export starts after a DTU reboot, its previous non-persistent commands are no longer known locally.

The controller therefore:

- Starts reachable, command-enabled inverters at 50 W
- Uses 50 W as the minimum production floor
- Waits for unreachable or night-disabled inverters until they become commandable
- Starts from nominal power when the previous operation ended with a 100% release

### Unreachable Inverters

When an inverter becomes temporarily unreachable:

- No new power-control command is queued for it
- Its last requested limit remains fixed
- Other reachable inverters absorb the correction when possible
- It is included again when communication returns

Release commands follow the same queue-safety rule. Commands are not sent to unreachable inverters and are retried when they become reachable.

## Grid Source Failure

If the selected grid source has no fresh value:

- Normal regulation is paused
- The previous grid value is not reused
- Reachable command-enabled inverters are moved to the 50 W fail-safe floor
- The fail-safe command is retried for individual unreachable inverters
- The non-persistent limit is periodically resent to prevent it expiring back to full power

When the grid source becomes available again, normal regulation resumes from the limits requested by Zero-Export.

MQTT publishes `unavailable` when the grid value is missing or expired, so Home Assistant does not keep displaying a stale retained value.

## Regulation Interval

The default regulation interval is 30 seconds.

This gives HMS inverters time to apply a new power limit and allows the grid source to measure and publish the resulting grid value before the next regulation cycle.

Once PR #3225 is merged, communication will become immediate and up to 10 inverters can be updated within approximately 2 seconds. The regulation interval can then safely be reduced to the configured minimum of 5-10 seconds.

## Configuration

The Zero-Export configuration includes:

- Enable/disable switch
- Grid power setpoint
- Minimum production power
- Regulation interval
- Grid source selection
- Shelly LNM multicast address
- Shelly LNM UDP port
- MQTT grid-power topic

The default minimum production is automatically calculated as:

```text
50 W x number of configured inverters
```

The configuration is included in the existing `config.json` export and import flow. Older configuration files remain compatible through defaults and migration handling.

## Web API

Added endpoints:

```text
GET  /api/zeroexport/config
POST /api/zeroexport/config
GET  /api/zeroexport/status
```

The status endpoint exposes:

- Zero-Export enabled state
- Setpoint
- Effective minimum production
- Regulation interval
- Selected source
- Grid power
- Grid power availability
- Current production limit

## Home Assistant

When MQTT Home Assistant discovery is enabled, OpenDTU publishes:

- Zero-Export enable switch
- Zero-Export setpoint number entity
- Grid Power sensor
- Production Limit sensor

The Grid Power and Production Limit entities are regular Home Assistant sensors, not diagnostic entities.

The enable switch and setpoint can be controlled directly from Home Assistant.

## Web Interface

Added a Zero-Export settings page under:

```text
Settings -> Zero-Export
```

The page provides:

- Enable/disable control
- Grid setpoint
- Minimum production power
- Regulation interval
- Grid source selection
- Shelly LNM configuration
- MQTT grid-power topic
- Live grid power
- Live production limit

Status values refresh automatically every 5 seconds.

The page is available in English, German, and French.

## MQTT and Radio Observability

This PR is based on PR #3226 and includes its command lifecycle logging.

Commands now have unique IDs that can be followed through:

```text
enqueue -> TX -> RX period end -> success/failure
```

Queue contents can also be inspected to diagnose:

- Commands blocked by unreachable inverters
- Retries and timeouts
- Dropped or replaced duplicate commands
- Delayed power-control commands
- Communication issues affecting the regulation loop

## Validation

The following checks pass:

- Firmware build for `generic_esp32s3_usb`
- Webapp production build
- Webapp Prettier check
- Configuration export/import integration
- `git diff --check`

## References

- Shelly Local Network Messaging: <https://shelly-api-docs.shelly.cloud/gen2/General/LocalNetworkMessaging/>
- Shelly LNM wire protocol: <https://shelly-api-docs.shelly.cloud/gen2/DynamicComponents/LNM/>
- Communication improvements: <https://github.com/tbnobody/OpenDTU/pull/3225>
- Command queue logging: <https://github.com/tbnobody/OpenDTU/pull/3226>

## Use Case

Automatic activation of Zero-Export or Full Export with 3ERL in France.

You can have a look at my Home Assistant Dashboard and how I now use OpenDTU + Shelly LNM to regulate my inverters:

https://gist.github.com/mathieucarbou/382556f1279d612962e03232544692d1

I simply need to switch the on/off flag of OpenDTU
And set an appropriate setpoint depending if my EV is plugged for charging or not.

### Shelly LNM Setup

Shelly LNM setup

### OpenDTU Zero-Export Configuration

**Shelly LNM:**

image

**MQTT**

Capture d’écran 2026-09-20 à 09 28 26

### Home Assistant Integration

Capture d’écran 2026-09-20 à 09 29 23

### Console Logs

Console logs

### Previous Home Assistant Regulation

Previous Home Assistant regulation

### Native OpenDTU Regulation

Native OpenDTU regulation

With PR #3225, regulation loop can be fastened and DTU Pol interval rduced:

655118428-7f79e9e6-420a-427b-9d7a-07e6a36feb7b

655118571-56e0721b-f221-4fe9-bc0e-df4067679600

## How try this firmware ?

You can find complete released firmware to download to test that from my fork:

=> https://github.com/mathieucarbou/tbnobody-OpenDTU/releases/

These firmware include these changes:

- **logging improvements**: https://github.com/tbnobody/OpenDTU/pull/3226
- **Zero-Export**: https://github.com/tbnobody/OpenDTU/pull/3227
- **Improve Inverter communication speed**: https://github.com/tbnobody/OpenDTU/pull/3225


mathieucarbou:
I have the regulation working since a few hours. Did some restarts, checked with some corner cases like inverter restart, unreachable, etc. Works great!

I will now test with the PR improving the communication speed: https://github.com/tbnobody/OpenDTU/pull/3225

mathieucarbou:
I tested with PR https://github.com/tbnobody/OpenDTU/pull/3225

## Regulation with only this PR:

Capture d’écran 2026-09-19 à 15 43 59

- DTU poll interval : 5 sec
- Regulation each 30 sec

With 8 inverters this is not possible top improve more.

## Including the fix in #3225

Capture d’écran 2026-09-19 à 16 17 16

image

image

Queue is drained quickly and power control are sent within 2 sec.

You can see in the screenshot that zero-export was deactivated and production power ramps up quickly.

mathieucarbou:
You can find complete released firmware to download to test that from my fork: 

**=> https://github.com/mathieucarbou/tbnobody-OpenDTU/releases/**

elmr91:
Thanks @mathieucarbou,

I'm currently testing your branch (from your repo including your 3 PRs) with MQTT measurement source.
It is working very well with 2xHMS-2000-4T
(BTW, I think I got my fist DTU-NRF24 from you few years ago)

My second HMS is a bit picky about timings and signal, but you branch is working fine.

I wonder if it would be possible to implement a dynamic limit relative to current production:
Limit max power to 20% above live production for example (when current production is too low).

Use case:
- All inverters are producing few power (under a cloud)
- It is not enough to reach the configured balance point
- Inverters limit is then slightly raised until the max 2 kW per inverter
- When cloud is gone, solar production is then suddenly raising far above the configured limit.
- It take some time to reduce the production.

We can also imagine to have steps (20, 40, 60, 80, 100 %) and set production limit to the next step above current production.
It may take 1 or more cycles to adjust to a new target if the target is very high, but is will work.

I tweaked the hoymileszeroexport from reserve85 this way ...
It was working fine.

mathieucarbou:
> I'm currently testing your branch (from your repo including your 3 PRs) with MQTT measurement source.
> It is working very well with 2xHMS-2000-4T

Thanks!

Do not forget that the regulation interval has to be configured depending on how fast the power control commands are dequeued. This can be seen in the logs. If you have an inverter relying on timeout and not early exit, then its command power control command could take 2 second to be dequeued.

In my case my HMS-1000-@R work nicely and are all applying power control within 2 sec.

> Limit max power to 20% above live production for example (when current production is too low).

It cannot be done easily because if you set an upper limit, it means you are allowing more excess. It is equivalent as setting a negative setpoint to increase the limit. The problem is that you never know if this allowance will lead to some production or not. If yes, then what do you do ? You step back ? You allow more excess ?

So if you play this game with trying to increase the limit always to "see" if it lead to production or not, just in order to place your limit above, you end up creating production and consumption spikes. That's not a good way to regulate.

The current algorithm as it is should be quite immediate because it does not work by reading inverter states.

It works by:
- reading grid power
- comparing it so setpoint
- spreading the difference across live inverters propotionnaly

So if you have a sudden production spike or decrease, the algorithm should adjust that in about 1-3 rounds max, so in my case 30 sec since I set a regulation interval of 10 sec.

elmr91:
This is the exact same algorithm I'm using.
The only difference is when calculating the new limit:

- reading grid power
- compare it to setpoint and calculate the new limit
- **if limit is over current produced power + 20% fullpower, adjust limit (for this round) to this value** 
- spread the difference across live inverters proportionaly

Inverters will increase their production (if possible), next round will adjust again if not enough to reach the setpoint.
In all other cases, it will not change anything to current behavior. 

The only idea is to avoid to have a limit way too high above what inverters are currently producing.
I already observed spikes of nearly 3 kW on my 4 kW system when weather is 50% cloudy (I'm also living in the "64").
I'm trying not sending too much to the network (they do not like spikes of injection especially over 3 kW)  

This will not change at all the global system behavior.

I'm trying to avoid such situation:
- current production : 500 W (big thick cloud)
- limit set to 2*2000 W
If direct sun reaches panels suddenly, system can increase production by few kW in seconds...  
If you set balance point to zero or some negative values, injection can easily go over this 3 kW limit.

mathieucarbou:
> Inverters will increase their production (if possible), next round will adjust again if not enough to reach the setpoint.
> In all other cases, it will not change anything to current behavior.

Oh! I think I did not understood what you meant by +20%.

If I am correct, you are stepping the increases and decreases to avoid sudden increases or decreases (spikes upon sudden changes) right ?

If so, this is equivalent to introducing a PID fo regulation.
Right now the regulation is a simple PID working on error where the I term would be 1.

It is definitely of course possible to add a PID implementation.

I did one for YaSolR (solar router - https://yasolr.carbou.me/manual/#pid) specially for MCU, that is quite tested and robust and even support a noise reduction feature, and it is in a lightweight lib that OpenDTU could definitely use (https://mathieu.carbou.me/MycilaUtilities/pid).

It even supports proportional on measurements, which helps limit overshoot (that's what you are talking about I think - avoiding overshoots).

But I doubt this would help.

here is my regulation today at ~ -200W since lunch time:

image

And a zoom on a spikes:

image

If you were working with a stepping PID, the grid power would not remain flat but would oscillate more because the sudden change would not be accounted for and would be progressive. So you end up having more excess or consumption.

With the changes did in OpenDTU with others PRs to fasten the com, you are not anymore subject to a slow regulation. Changes are applies immediately in less than 2 seconds. Shelly LNM allows a 1 sec interval data: this is even more precise than what HA gets from MQTT because Shelly's only do MQTT publications when the % change in power  is above a threshold. With Shelly LNM this is different: every state is received each second.

And the regulation is fast: each 10 sec in my case but could even be lowered.

So with such second-level scale, there is no need to step the increase or decrease I think...  
Especially also because grid providers are monitoring zero-export not on a second scale but on a minute scale.

This would need more testing of course, I can be wrong, but let's see if these cases happen first before introducing a PID.

For example we see in the graph such spike where the grid power bumped to +1000W then -1000W. But these spikes lasted what ? about 10 seconds. Compared to if you were using a PID which would adjust in 2-3 rounds,  you would probably consume or inject more maybe ?

elmr91:
I'm using MQTT because my Shelly is Gen1 without LNM support.

For variations within +20% of fullpower (maybe around +1kW in your case) the proposed algorithm would have produced the exact same curve you got (when variations remained within +1 kW range):
image

There would have been slight changes when increasing quickly production:
It would have taken 1 or 2 intermediate steps instead of only one big step:
image


Note : I'm fighting with one of my inverters that sometimes does not want to respond at all (or only 66% sucess).
It is difficult to test today -> It is in a bad mood day.

mathieucarbou:
> I'm using MQTT because my Shelly is Gen1 without LNM support.

Ah! Indeed! Look at what is happening with MQTT grid power reported in HA:

- in red, all the Shelly values pushed by OpenDTU through MQTT each 5 sec (we do not have better granularity with MQTT publication in OpenDTU sadly)

- in cyan: all the received grid power from MQTT pushed by Shelly

image

MQTT is slower and misses a lot of values, especially where there is a gradual increase or a short pulse.

image

This spike was seen by OpenDTU who increased production but this was a short spike so then just after decreased production. But it took more time for the Shelly to publish a MQTT update.

**So yes, MQTT being slower, in this case, you would definitely benefit of a PID that is ramping up the correction.**

=> I can definitely add a variant of that if you are ready to test ? Otherwise if it was just a feature request or idea, just let me know when you have everything setup and ready to test so that you could have at least a before and after comparison. Implementing that is quick.

> Note : I'm fighting with one of my inverters that sometimes does not want to respond at all (or only 66% sucess).

I got the same issue with 2 of mine. I went on the roof and discovered that we did not pay attention to the inverter orientation when installing them. Them have a short radio but so many obstacles... I just moved the inverter 1 meter on the roof and installed in the correct orientation for the antenna to have less obstacles.

In my case this was a stupid installation mistake... We did not account for radio com and obstacles.

elmr91:
I would be very happy to test and experiment with a tuned regulation.

Shelly Gen1 behave better than new gen regarding MQTT updates.
It can update every second if measurement changes (by few %).
If measurement is still, it updates every 30s maximum (I changed it to 10s max between 2 reports) 


Initially my HMS inverters were both working fine.
They are installed under the panels on a nearly flat (10%) iron roof (bac acier).
They both handle a row of 4 panels.

They are installed the same way, antenna on the top.
DTU is 10m away (behind a concrete wall) : -61 dBm for the bottom inverters (HMS1), -68 dBm for the top inverters (HMS2)
I would have expected rather the same signal level for both.
I already lost communication with HMS2 several times.
I think something happened to this one, it has degraded over time.

When it works well, it always has lowest RX success (around 66-75%), the other is 100%

Edit: 
I have juste moved my DTU just below the roof where inverters are installed Signal increased to -46/-51 dBm and 2 MHS are online with 100% RX success.
But now WiFi signal is too weak.
I may have to rebuild my DTU with an ESP+external Wi-Fi antenna.

mathieucarbou:
> I would be very happy to test and experiment with a tuned regulation.

I will make up something and let you know. My grid MQTT updates are worse than yours so if the solution works for me it will for you ;-)

There were some appliances stop/start at home and see the mqtt update saw nothing.

image

> Shelly Gen1 behave better than new gen regarding MQTT updates.
> It can update every second if measurement changes (by few %).

Didn't know!

> If measurement is still, it updates every 30s maximum (I changed it to 10s max between 2 reports)

Pay attention to set a regulation interval in OpenDTU that is more than the frequency at which the measurements are coming. Because if you do not do that, **you can end up with the same difference between last grid power and setpoint that will be re-applied again**, and this causes an over-consumption or excess.

The curent implementation is the equivalent of an integral-only controller with gain 1. 
**It's unconditionally stable only if the regulation period ≫ source latency + inverter response.** 

- With Shelly LNM (1 s updates), a 5–30 s interval gives ~3–10 fresh samples per correction — fine.

- **With a slow MQTT source (say 30 s updates) and a 5 s interval, you need to have a loop interval in zero-export config that is at leats 35 seconds.**

elmr91:
DTU moved very near from the inverters.

WiFi is in range (low but usable) / MQTT updates are coming quickly when noticeable changes occurred :
image


- Regulation loop set to 15s
- Shelly/MQTT set to update at least every 10s (or lower in case of change)
- Early exit is triggered at every limit change

image

mathieucarbou:
> DTU moved very near from the inverters.
> 
> * Regulation loop set to 15s
> * Early exit is triggered at every limit change
> 

is that better now for the graphs ?

I am currently doing an update to keep track of the last update intervals in an exponential moving average in order to determine the average update interval time for the metrics and consequently provide a limit update that is in function of that. For example, if the calculated average delay is 10 sec to get a new grid power update and the regulation interval is 5 second, then when applying a new power control, half of the value will be applied (something similar).

This is a source-adaptive correction which is automatically spreading the correction to apply across the time.

The advantage of that: user does not need to know that the interval of the regulation has to be larger than the delays at which the grid power is coming.

elmr91:
Example of scenario were home consumption if higher than solar production (+setpoint)

Stable starting point, stable home consumption, setpoint -800W
Solar is producing and limited to 1600W. 

```
[16:31:26.517] I (2467420) zeroexport: Grid -796.7 W, setpoint -800.0 W -> adjusting total limit from 1600 to 1604 W
[16:31:31.673] D (2472579) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:31:32.172] I (2472579) zeroexport: Grid Power from MQTT: -815.3 W
[16:31:35.565] D (2476472) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:31:36.524] I (2476473) zeroexport: Grid Power from MQTT: -843.9 W
[16:31:36.524] I (2477420) zeroexport: Grid -843.9 W, setpoint -800.0 W -> adjusting total limit from 1604 to 1561 W
[16:31:39.470] D (2480377) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:31:40.885] I (2480378) zeroexport: Grid Power from MQTT: -792.2 W
[16:31:40.885] D (2481793) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:31:42.177] I (2481793) zeroexport: Grid Power from MQTT: -795.7 W
[16:31:42.278] D (2483132) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)

```

Starting a 1200W device.
Home consumption : -800W -> 400W

```
[16:31:42.278] I (2483133) zeroexport: Grid Power from MQTT: 379.1 W
[16:31:43.447] D (2484355) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:45.001] I (2484356) zeroexport: Grid Power from MQTT: 359.2 W
```

Regulation increased limit by nearly 1200W : OK

```
[16:31:46.515] I (2487420) zeroexport: Grid 359.2 W, setpoint -800.0 W -> adjusting total limit from 1562 to 2722 W
[16:31:47.277] D (2488120) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:47.277] I (2488120) zeroexport: Grid Power from MQTT: -75.6 W
[16:31:50.201] D (2491110) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:51.336] I (2491111) zeroexport: Grid Power from MQTT: -72.8 W
[16:31:51.336] D (2491219) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:51.336] I (2491220) zeroexport: Grid Power from MQTT: -69.4 W
[16:31:51.336] D (2492236) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:52.188] I (2492237) zeroexport: Grid Power from MQTT: -84.7 W
[16:31:53.480] D (2494388) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:56.517] I (2494389) zeroexport: Grid Power from MQTT: -85.4 W
```

But there is not enough solar power to reach the limit
Regulation tries to increase again the limit
This is one case where limiting max power above xx% of current production could make sense   

```
[16:31:56.517] I (2497420) zeroexport: Grid -85.4 W, setpoint -800.0 W -> adjusting total limit from 2722 to 3437 W
[16:31:58.308] D (2499218) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:31:59.932] I (2499219) zeroexport: Grid Power from MQTT: -30.9 W
[16:31:59.933] D (2500841) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:00.443] I (2500842) zeroexport: Grid Power from MQTT: -30.9 W
[16:32:00.443] D (2501353) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:01.387] I (2501353) zeroexport: Grid Power from MQTT: -79.8 W
[16:32:01.387] D (2502296) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:02.188] I (2502296) zeroexport: Grid Power from MQTT: -96.5 W
[16:32:03.413] D (2504322) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:05.768] I (2504322) zeroexport: Grid Power from MQTT: -80.8 W
[16:32:05.768] D (2506676) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:32:06.517] I (2506677) zeroexport: Grid Power from MQTT: -108.4 W
```

Once again regulation increased limit, reaching the max
(I cut a bit of logs below)


```
[16:32:06.517] I (2507420) zeroexport: Grid -108.4 W, setpoint -800.0 W -> adjusting total limit from 3438 to 4000 W
[16:32:07.289] D (2508119) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:32:07.289] I (2508120) zeroexport: Grid Power from MQTT: -107.7 W
[16:32:08.235] D (2509142) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:32:12.194] I (2509142) zeroexport: Grid Power from MQTT: -107.7 W
[16:32:12.292] D (2513122) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:12.292] I (2513124) zeroexport: Grid Power from MQTT: -92.6 W
[16:32:13.347] D (2514255) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:32:15.703] I (2514257) zeroexport: Grid Power from MQTT: -102.7 W
[16:32:15.703] D (2516611) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:16.419] I (2516612) zeroexport: Grid Power from MQTT: -37.5 W
[16:32:16.419] D (2517326) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:17.198] I (2517327) zeroexport: Grid Power from MQTT: -19.5 W
[16:32:17.198] D (2517421) zeroexport: Grid -19.5 W, setpoint -800.0 W -> no adjustment needed (total limit 4000 W)
[16:32:17.198] D (2517551) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:17.198] I (2517552) zeroexport: Grid Power from MQTT: -19.5 W
[16:32:17.294] D (2518127) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-4/4)
[16:32:17.294] I (2518128) zeroexport: Grid Power from MQTT: 1.9 W
[16:32:19.204] D (2520111) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:32:22.200] I (2520112) zeroexport: Grid Power from MQTT: -8.0 W
[16:32:23.503] D (2524411) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:26.511] I (2524411) zeroexport: Grid Power from MQTT: -16.3 W
[16:32:26.511] D (2527420) zeroexport: Grid -16.3 W, setpoint -800.0 W -> no adjustment needed (total limit 4000 W)
[16:32:28.094] D (2529003) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-4/4)
[16:32:28.149] I (2529004) zeroexport: Grid Power from MQTT: 2.0 W
[16:32:28.149] D (2529004) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-6/6)
[16:32:28.149] I (2529005) zeroexport: Grid Power from MQTT: -10.2 W
[16:32:28.149] D (2529033) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:32:28.149] I (2529034) zeroexport: Grid Power from MQTT: -8.2 W
[16:32:28.149] D (2529035) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:32:28.149] I (2529035) zeroexport: Grid Power from MQTT: -8.2 W
[16:32:29.659] D (2529035) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:32:29.659] I (2529035) zeroexport: Grid Power from MQTT: -3.0 W
[16:32:29.659] D (2530474) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-4/4)
[16:32:30.119] I (2530477) zeroexport: Grid Power from MQTT: 10.0 W
[16:32:31.581] D (2531095) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-4/4)
[16:32:31.581] I (2531095) zeroexport: Grid Power from MQTT: 6.0 W
[16:32:31.581] D (2532490) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:32:32.205] I (2532490) zeroexport: Grid Power from MQTT: 10.7 W
[16:32:35.669] D (2536579) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:32:36.511] I (2536579) zeroexport: Grid Power from MQTT: 14.7 W
[16:32:36.512] D (2536798) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-4/4)
[16:32:36.512] I (2536799) zeroexport: Grid Power from MQTT: 0.0 W

************

[16:33:26.510] D (2587420) zeroexport: Grid 4.7 W, setpoint -800.0 W -> no adjustment needed (total limit 4000 W)
[16:33:27.350] D (2588185) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:33:27.350] I (2588187) zeroexport: Grid Power from MQTT: 16.9 W
[16:33:28.678] D (2589575) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:33:28.978] I (2589577) zeroexport: Grid Power from MQTT: 34.4 W
[16:33:28.978] D (2589658) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-5/5)
[16:33:29.436] I (2589658) zeroexport: Grid Power from MQTT: 34.4 W
[16:33:29.436] D (2590346) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)

```
I stopped the 1200W load.
Production is now a bit above -800W setpoint.
Regulation tries to decrease production by lowering the limit
As limit was way too high (max 4000W) above the real solar production, it take many cycles for the limit to take effect.

```
[16:33:31.480] I (2590347) zeroexport: Grid Power from MQTT: -1096.6 W
[16:33:31.480] D (2592391) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:33:32.255] I (2592391) zeroexport: Grid Power from MQTT: -1120.8 W
[16:33:36.512] I (2597420) zeroexport: Grid -1120.8 W, setpoint -800.0 W -> adjusting total limit from 4000 to 3680 W
[16:33:38.135] D (2599045) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:33:38.736] I (2599045) zeroexport: Grid Power from MQTT: -1094.5 W
[16:33:46.514] I (2607420) zeroexport: Grid -1094.5 W, setpoint -800.0 W -> adjusting total limit from 3680 to 3386 W
[16:33:47.266] D (2607675) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:33:47.266] I (2607676) zeroexport: Grid Power from MQTT: -1099.3 W
[16:33:56.039] D (2616951) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:33:56.512] I (2616952) zeroexport: Grid Power from MQTT: -1128.7 W
[16:33:56.512] I (2617420) zeroexport: Grid -1128.7 W, setpoint -800.0 W -> adjusting total limit from 3386 to 3058 W
[16:34:04.883] D (2625795) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:34:06.512] I (2625796) zeroexport: Grid Power from MQTT: -1110.5 W
[16:34:06.512] I (2627420) zeroexport: Grid -1110.5 W, setpoint -800.0 W -> adjusting total limit from 3058 to 2748 W
[16:34:13.857] D (2634761) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:34:16.512] I (2634763) zeroexport: Grid Power from MQTT: -1102.1 W
[16:34:16.512] I (2637420) zeroexport: Grid -1102.1 W, setpoint -800.0 W -> adjusting total limit from 2748 to 2446 W
[16:34:22.974] D (2643887) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:34:26.515] I (2643889) zeroexport: Grid Power from MQTT: -1070.2 W
[16:34:26.515] I (2647420) zeroexport: Grid -1070.2 W, setpoint -800.0 W -> adjusting total limit from 2446 to 2176 W
[16:34:32.002] D (2652911) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:34:32.292] I (2652911) zeroexport: Grid Power from MQTT: -1057.5 W
[16:34:33.431] D (2654344) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:34:36.512] I (2654344) zeroexport: Grid Power from MQTT: -1057.1 W
[16:34:36.512] I (2657420) zeroexport: Grid -1057.1 W, setpoint -800.0 W -> adjusting total limit from 2176 to 1919 W
[16:34:41.053] D (2661963) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-8/8)
[16:34:42.301] I (2661964) zeroexport: Grid Power from MQTT: -1032.7 W
[16:34:46.514] I (2667420) zeroexport: Grid -1032.7 W, setpoint -800.0 W -> adjusting total limit from 1920 to 1688 W
[16:34:47.403] D (2668234) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:34:47.403] I (2668235) zeroexport: Grid Power from MQTT: -829.0 W
[16:34:49.187] D (2670097) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:34:49.815] I (2670098) zeroexport: Grid Power from MQTT: -781.1 W
[16:34:49.815] D (2670731) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:34:52.304] I (2670731) zeroexport: Grid Power from MQTT: -781.1 W
[16:34:56.512] I (2677420) zeroexport: Grid -781.1 W, setpoint -800.0 W -> adjusting total limit from 1688 to 1707 W
[16:35:00.178] D (2680773) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:35:02.312] I (2680775) zeroexport: Grid Power from MQTT: -806.5 W
[16:35:06.511] I (2687420) zeroexport: Grid -806.5 W, setpoint -800.0 W -> adjusting total limit from 1708 to 1702 W
[16:35:12.317] D (2688779) mqtt: Received MQTT message on topic 'shellies/shellyem-E868E7D2B0B5/emeter/0/power' (Bytes 1-7/7)
[16:35:12.317] I (2688779) zeroexport: Grid Power from MQTT: -793.1 W
```

elmr91:
The algorithm is OK as soon as solar is producing enough.
It works less well in under production cases if we let the limit going too far from the effective production.

mathieucarbou:
> The algorithm is OK as soon as solar is producing enough. It works less well in under production cases if we let the limit going too far from the effective production.

I am testing the new algorithm that is auto-correcting the new limites based on the EMA of the measurement delays.

image

I still have the regulation loop running each 10 sec, but this time, with my infrequent grid power updates.

You can see that the algorithm now paces the increases or decreases (steps in graphs) based on the delays these updates were received.

If the regulation loop is faster than the measurement delays, some warnings are shown in the logs to display the % of correction applied:

```
[17:08:04.958] W (716386) zeroexport: Slow grid source (period 23478 ms): correction factor: 39%
[17:08:14.943] W (726386) zeroexport: Slow grid source (period 21052 ms): correction factor: 43%
[17:08:24.940] W (736386) zeroexport: Slow grid source (period 21052 ms): correction factor: 43%
[17:08:34.943] W (746386) zeroexport: Slow grid source (period 21052 ms): correction factor: 43%
[17:08:44.976] W (756386) zeroexport: Slow grid source (period 21052 ms): correction factor: 43%
[17:08:54.939] W (766386) zeroexport: Slow grid source (period 21052 ms): correction factor: 43%
[17:09:02.558] W (773654) zeroexport: Zero-Export MQTT payload on homeassistant/states/sensor/grid_power/state could not be parsed as grid power
[17:09:14.948] W (786387) zeroexport: Slow grid source (period 20850 ms): correction factor: 44%
[17:09:25.440] W (796886) zeroexport: Slow grid source (period 17012 ms): correction factor: 53%
[17:09:55.439] W (826886) zeroexport: Slow grid source (period 9480 ms): correction factor: 87%
[17:10:15.477] W (846256) hoymiles: All missing
[17:10:15.477] W (846256) hoymiles: Nothing received, resend whole request: [1194] RealTimeRunData (1410a01e2cdd)
[17:11:08.313] W (899767) zeroexport: Zero-Export MQTT payload on homeassistant/states/sensor/grid_power/state could not be parsed as grid power
[17:11:15.940] W (907386) zeroexport: Slow grid source (period 12184 ms): correction factor: 71%
[17:11:25.978] W (917386) zeroexport: Slow grid source (period 10387 ms): correction factor: 81%
[17:11:38.721] W (929522) hoymiles: All missing
[17:11:38.721] W (929522) hoymiles: Nothing received, resend whole request: [1315] ActivePowerControl (HMS, 00) (1410a01f010c)
[17:12:21.385] W (972831) zeroexport: Zero-Export MQTT payload on homeassistant/states/sensor/grid_power/state could not be parsed as grid power
[17:12:26.444] W (977886) zeroexport: Slow grid source (period 13970 ms): correction factor: 63%
[17:12:36.444] W (987886) zeroexport: Slow grid source (period 13970 ms): correction factor: 63%
```

I discovered that besides being infrequent, HA also published unavailable sometimes lol in case Shelly does not update in time 😆 I really have the poorest MQTT use case lol so this should be better on your side.

mathieucarbou:
Works fine with a source-adaptive correction:

image

and this is on my poor MQTT setup ;-)

mathieucarbou:
@elmr91 : you can install [v26.3.30-mathieu8](https://github.com/mathieucarbou/tbnobody-OpenDTU/releases/tag/v26.3.30-mathieu8) ;-) I did also and will keep it running tomorrow also to see how it behaves with bigger load spikes.

Works fine with a source-adaptive correction:

image

and this is on my poor MQTT setup ;-)

elmr91:
I have installed the latest beta few minutes ago.

Here are some graphs from this afternoon (previous beta)
image
We can see AC power following production limit by few seconds. 

Load spike (17:43) has been well handled -> 5 seconds measured before increasing power and back to setpoint -800W
image

Problems are only when production is too low to reach setpoint -> limit is slightly increasing up to the max.
This is where I want to stop the raising of the limit (to some amount above the current production)
image

mathieucarbou:
image

 This is what I get at the end of the day.
A lot increases also in my case so production limit goes up until the max, then load stops and limit goes down.

I agree with you that having a way to cap the increase might be nicer for the graph but as for the regulation system in my case there would not be any difference since going down is fast.

you have exactly the same behavior: ramping up is slow but going down is fast.

So what would be the benefit of hacking into the regulation system to set a corner cases and cap it when it is anyway already capped at the max production limit ?

elmr91:
There are  2 cases were capping the limit is useful:

- preventing too high production spikes when insultation suddenly increased (cloudy weather)
- speedup production decrease on small load decrease (in case of under production)

Example for second case:
- balanced state with just a little under production: setpoint = -800w, measurement = -700w , solar production = 1000 W, limit raised to 4000 W (under production) 
- stop a 300W load, measurement raised to -1000 W, eg. -200W over setpoint.
- limit will be decreased by 200W every round.
- It will take (4000-1000)/200 -> 15 rounds to decrease the solar production
(it will be quicker for larger loads but will still be noticeable) 

If limit is kept a little above production, it will take less rounds.

I agree the system converges quicker with larger load changes.
If load changes per 1 or 2 kW, only few rounds will be required.

mathieucarbou:
I will try to find a solution. 
what is bothering me more is why I still see the limit going up and down even when the solar production is below the home consumption now…

image

elmr91:
I don't have the same behavior with beta8 (no limit decrease):
image

May be are you testing some additional tweaks...
from 20:30, my battery is taking over -> zero consumption target.

mathieucarbou:
> I don't have the same behavior with beta8 (no limit decrease): image
> 
> May be are you testing some additional tweaks... from 20:30, my battery is taking over -> zero consumption target.

There is a failsafe mechanism that ensures that if Zero-Export does not receive any grid power update after a time (60 sec for mqtt and 10 sec for Shelly LNM), then the algo is automatically entering in failsafe mode and pushes a limit to 50W on each inverter each 10min because some inverters are resetting to 100% if not updated after a while.

So this failsafe mechanism is triggerring for me because the grid power does not update for more than 1 min.

image

mathieucarbou:
@elmr91 : I was able to make it work:

Firmware: https://github.com/mathieucarbou/tbnobody-OpenDTU/releases

### Production Cap

I introduced what I call a production cap which is computed by looking at the current DTU total AC production and by adding a margin to it which is either 200W or 20% of the production (whatever the max is).

So when we are rising the power limits to allow more production, this limit is enforced to cap the increase.

But when decreasing the power limits because the production is too high, then this cap is not considered because we want to go below the current solar production.

image

setpoint at -200W since 9:40 am

### Failsafe Timeout

For each grid source a failsafe timeout is now configurable.

Capture d’écran 2026-09-21 à 07 33 48

For example in my case my Home Assistant grid power can be left without any update for a long time, sometimes more than 2 minutes. Once this timeout is expired without any new value coming, inverters are put in failsafe mode at their minima production power.

### Minimal Production Power per inverter

It was changed to be a configuration per inverter with a minimal threshold of 40W.

Capture d’écran 2026-09-21 à 07 44 32

elmr91:
Thanks a lot @mathieucarbou ,
I have upgraded to the latest beta.

My battery is charging in the morning, "eating" all solar production in excess. Inverters are producing at max power:
1000W with 1200W high limit -> Cap is working fine !

image


Battery will be full within 30mn, I will be able to monitor how regulation behave when inverters power would be limited.

mathieucarbou:
> Thanks a lot @mathieucarbou , I have upgraded to the latest beta.
> 
> My battery is charging in the morning, "eating" all solar production in excess. Inverters are producing at max power: 1000W with 1200W high limit -> Cap is working fine !
> 
> image
> Battery will be full within 30mn, I will be able to monitor how regulation behave when inverters power would be limited.

Ok. FYI I saw another existing bug in OpenDTU in the com layer(c++ programming mistake).

This bug (deadlock) is triggered if the grid power stops arriving after a while (after the failsafe timeout). 
Then,, the DTU keeps sending each 10min some power limit controls with the minimal value. 
At one point, it will deadlock and need to be restarted (power shutdown).

I will send a separate PR for this issue and integrate the fix in a release soon.

elmr91:
Focus on reaction time:
image
Red line is Shelly measurement seems by HA (MQTT)

- Limit started to decrease 10s later (10s loop)
- Inverters started to decrease their production 2s later (Shelly curve)
- AC power (reported by DTU) started to decrease 5s later (delayed due to 10s DTU polling I guess)
- Target production reached 30s after the initial event. 

It seems to be coherent with your design.

May be we could raise the decrease slope ?
Currently 2 rounds to reach the final target. Is this due to your new regulation algorithm ?
The initial (simpler) algorithm would have reached to target in only 1 round.



Edit: another one in the other direction -> increasing load +1200W
image
40 seconds to reach target: 3 rounds of (2*200 W) 
4 rounds in total (including initial one).

mathieucarbou:
> Currently 2 rounds to reach the final target. Is this due to your new regulation algorithm ?

Yes thats might be because your regulation interval is lower than the time between the incoming mqtt values.

elmr91:
> > Currently 2 rounds to reach the final target. Is this due to your new regulation algorithm ?
> 
> Yes thats might be because your regulation interval is lower than the time between the incoming mqtt values.

This is really strange...
Red line represents MQTT updates from the Shelly sensor received directly by HA

I pointed all MQTT value changes. They are really more frequent than regulation loop duration (10s).
image

mathieucarbou:
> > > Currently 2 rounds to reach the final target. Is this due to your new regulation algorithm ?
> > 
> > 
> > Yes thats might be because your regulation interval is lower than the time between the incoming mqtt values.
> 
> This is really strange... Red line represents MQTT updates from the Shelly sensor received directly by HA
> 
> I pointed all MQTT value changes. They are really more frequent than regulation loop duration (10s). image

don't forget you also have the 5 sec interval mqtt update of opendtu and the regulation interval so the values seen by HA are seen by opendtu also (same publication time), but the DTU can wait before using it and publishing it.

elmr91:
> > > > Currently 2 rounds to reach the final target. Is this due to your new regulation algorithm ?
> > > 
> > > 
> > > Yes thats might be because your regulation interval is lower than the time between the incoming mqtt values.
> > 
> > 
> > This is really strange... Red line represents MQTT updates from the Shelly sensor received directly by HA
> > I pointed all MQTT value changes. They are really more frequent than regulation loop duration (10s). image
> 
> don't forget you also have the 5 sec interval mqtt update of opendtu and the regulation interval so the values seen by HA are seen by opendtu also (same publication time), but the DTU can wait before using it and publishing it.

This is why the orange line is also there on the graph: DTU taking input from MQTT sensor and pushing back it its own MQTT topic "Export grid power".

I'll try to test again int the coming days while getting the console log to better understand what is happening.

mathieucarbou:
> > > > > Currently 2 rounds to reach the final target. Is this due to your new regulation algorithm ?
> > > > 
> > > > 
> > > > Yes thats might be because your regulation interval is lower than the time between the incoming mqtt values.
> > > 
> > > 
> > > This is really strange... Red line represents MQTT updates from the Shelly sensor received directly by HA
> > > I pointed all MQTT value changes. They are really more frequent than regulation loop duration (10s). image
> > 
> > 
> > don't forget you also have the 5 sec interval mqtt update of opendtu and the regulation interval so the values seen by HA are seen by opendtu also (same publication time), but the DTU can wait before using it and publishing it.
> 
> This is why the orange line is also there on the graph: DTU taking input from MQTT sensor and pushing back it its own MQTT topic "Export grid power".
> 
> I'll try to test again int the coming days while getting the console log to better understand what is happening.

Don't rely on "Zero Export Grid Power" since it will be removed. This is a data that comes from elsewhere to OpenDTU so this is not OpenDTU responsibility to expose it, and besides, it exposes it later.

I was not able to test today: still figuring out my deadlock issue. It happens because my MQTT is too slow and it deadlocks at some point, and/or this is the MQTT calls that triggers the setpoint changes I don't know.

mathieucarbou:
> I was not able to test today: still figuring out my deadlock issue. It happens because my MQTT is too slow and it deadlocks at some point, and/or this is the MQTT calls that triggers the setpoint changes I don't know.

That's this bug: https://github.com/tbnobody/OpenDTU/pull/3217

So deadlock was happening when a modification of config was incoming through mqtt.

mathieucarbou:
# OpenDTU Zero-Export - Update

**Firmware location to test**

=> https://github.com/mathieucarbou/tbnobody-OpenDTU/releases

## Breaking Changes

- `Minimal production Power` setting is per inverter! If you have changed this vlaue, be sure to update it!
- `Zero-Export Grid Power` is removed from Home Assistant integration
- Changing the setpoint from MQTT is transient and not persisted in configuration file. This is to avoid wear and tear and avoid temporary automation states to be saved.
- Changing the zero-export switch from MQTT is transient and not persisted in configuration file. This is to avoid wear and tear and avoid temporary automation states to be saved.

## Changes

- API `/api/zeroexport/config`
- API `/api/zeroexport/status`
- Failsafe timeout is configurable per grid source
- `Zero-Export Grid Update interval` is shown in UI
- `Zero-Export Grid Update Interval` is added to Home Assistant integration
- `Zero-Export Failsafe` is added to Home Assistant integration (activated when OpenDTU is not receiving the grid power anymore)
- Several deadlocks fixed
- Updated constants to raise the production limit cap to +300W / 30% of current one

## Pull Requests

### Included

**fix: WriteGuard deadlock**
=> https://github.com/tbnobody/OpenDTU/pull/3217
(08c42fa327ffd57ddfb47457503ac9880567a86d)

**fix(hoymiles): guard empty command queue during deduplication**
=> https://github.com/tbnobody/OpenDTU/pull/3231
(379d501fa662885727435aa76fe69e24a350b8a9)

**fix: console log lines misclassified as warnings**
=> https://github.com/tbnobody/OpenDTU/pull/3230
(d4c41d9725b10d05dfe245872c1d6725cac66e50)

**feat(hoymiles): early RX exit for faster inverter control**
=> https://github.com/tbnobody/OpenDTU/pull/3225
(143318432ee70eb733bf3184466b60dbe7d53190)

**feat: OpenDTU Zero-Export (supports Shelly LNM or MQTT)**
=> https://github.com/tbnobody/OpenDTU/pull/3227

### Not Included

**feat(hoymiles): track commands through the radio queue with unique IDs**
=> https://github.com/tbnobody/OpenDTU/pull/3226
(f5bde17915a9280774598e4770abb51f89214dcc)

## Screenshots

Capture d’écran 2026-09-22 à 11 04 57

Capture d’écran 2026-09-22 à 11 06 01

Capture d’écran 2026-09-22 à 11 06 10

Capture d’écran 2026-09-22 à 11 06 29

mathieucarbou:
@elmr91 : FYI you can install the new release and report your findings once tested.

I am exposing through HA the grid update interval.

It is really really poor in my case. So I am testing with the worst possible conditions today: regulation interval 30 sec and my shelly updating grid power each 10-20  sec or more sometimes.

Tomorrow I will test with Shelly LNM measures arriving each 1 sec, regulation each 10sec.

image

Here is the production starting this morning:

image

Here is now:

image

Grid power of course fluctuates a lot because of my poor regulation condition (bad MQTT refresh + 30 sec regulation). 
There is nothing that can be done to improve a slow measurement system because it will be subject to oscillations.

elmr91:
Thanks Mathieu,

I have upgrade to beta10 (remotely from office today).
It is running fine since one hour.

I got between 5-7s "Zero-Export Grid Update Interval" reported by DTU:
(never below 5s)

image

It is far above MQTT updates pushed by Shelly.
I think DTU is only taking MQTT events every 5s.


FYI, I had to manually set DTU settings after every boot: https://github.com/tbnobody/OpenDTU/pull/3101
In main branch, PAlevel is ignored (overwritten) at boot until it is manually set using GUI (DTU settings).
As I put my DTU very near from inverters, signal is too strong. I need to set -7 dBm to get 100% RX success.

mathieucarbou:
> FYI, I had to manually set DTU settings after every boot: #3101

I will include this fix in my builds then! You confirm it fixes your issue taken as-is ?

elmr91:
> > FYI, I had to manually set DTU settings after every boot: #3101
> 
> I will include this fix in my builds then! You confirm it fixes your issue taken as-is ?

I'll test a build with this specific PR to be 100% sure it fixed the problem.
Easy to confirm in my current setup:

- Reboot a DTU build without the fix 33% success RX
100% RX success after setting DTU parameters
- Reboot a DTU build with the correct fix should provide 100% success RX

mathieucarbou:
> > > FYI, I had to manually set DTU settings after every boot: #3101
> > 
> > 
> > I will include this fix in my builds then! You confirm it fixes your issue taken as-is ?
> 
> I'll test a build with this specific PR to be 100% sure it fixed the problem. Easy to confirm in my current setup:
> 
> * Reboot a DTU build without the fix 33% success RX
>   100% RX success after setting DTU parameters
> * Reboot a DTU build with the correct fix should provide 100% success RX

I released a new build to add this PR: [v26.3.30-mathieu12](https://github.com/mathieucarbou/tbnobody-OpenDTU/releases/tag/v26.3.30-mathieu12).

I also flashed it, in order to test it tomorrow.


FYI here is my day today.

This was mostly to test the starting and ending with a slow MQTT use case and also how it behaves during load spikes.
My MQTT update interval can be really high... nearly 1 min at the end!

Someone having a more frequent and stable QTT update will definitely have a better graph!

What is interesting is the end: once the solar production is not enough, the limit is raised but up to +300W or +30% max.

image

image

But even with these spikes the energy graph looks acceptable for someone who is not picky as me and is OK with a slow regulation system...

image

 **I have exported 1.73kWh with a setpoint at -200W over 10 hours, so that's pretty good!**

Tomorrow I will test with Shelly LNM / 10 sec regulation and within the day with the EV car charging to see how it behaves to quickly ramp up production when the EV car connects.

I expect a beautiful regulation tomorrow!

elmr91:
Regulation result this afternoon with beta10:
image
Unfortunately not much power spikes...

Just this ones : very short power spikes.
When regulation reacts, the load is nearly already over. Regulation goes in the other direction the next rounds. This is an expected behavior.
image

I'll try to generate tomorrow (working from home) more load changes for testing.

elmr91:
It may be useful to implement a dead zone or hysteresis in order to avoid too small changes in the limit setpoint
Updates for -/+2W are often pushed to inverters:
image

I'm not sure inverters can regulate production so precisely.

mathieucarbou:
> t may be useful to implement a dead zone or hysteresis in order to avoid too small changes in the limit setpoint
> Updates for -/+2W are often pushed to inverters:

there is one currently, the code supports that, but it is not configurable currently and only prevents if less than 1W adjustment.

what is complicated is that this dead zone acts on the total production which has to be spread across inverters after. If you have a mix of HMS-2000-4T and HMS-1000-2T you will have your 4T have twice the change compared to the 2T.

So yes, 1W does not make a difference. But someone having 20 inverters with a mix of 50/50 2T and 4T, 1W on a 2T means 2W on a 4T which means a production change of 30W.

30W for such installation is low. But 30W for someone with only 1 inverter is high.

So I prefer for now to keep things as simple as possible without too much complexity.

We'll see after if we need to tweak this constant or make it configurable.

**FYI: I have released version 12 to include another fix in the com stack.**

elmr91:
Upgraded to beta12:

I confirm https://github.com/tbnobody/OpenDTU/pull/3101 is working perfectly in your build.
No need to set DTU at every boot anymore : PAlevel are correctly applied on boot. 

I had a look on Grid Update interval: this morning it is around 1s (as I expected):
image

Updates are coming every 1s (more or less):
image

Did you change anything from previous build ? (reported interval was above 5s)

Edit: it seems reporting interval seems fluctuating (very strange):
image
I'll monitor MQTT event frequency. May be related to my Shelly (I'm thinking to upgrade it to a gen3)

elmr91:
Regarding current dead zone, it seems very small.
There are +/-1W limit adjustments in logs (I checked powerControl commands are sent to inverters at every limit change): 

image

mathieucarbou:
> Did you change anything from previous build ? (reported interval was above 5s)

Nothing regarding this part. Note that the interval returned is not the last one but an EMA based on previous values in order to damp potential spikes. So if it raises a lot it means you had several mqtt publish that were slower.

It could be:

1) opendtu slowdown
2) HA publish slowdown

For the first case, which is highly likely, opendtu can slwodown sometimes in com stack or processing stuff. But not up to 8 sec like you had. You would need to find a way to monitor that in HA maybe.

Here is my start of the day with Shelly LNM and 10 sec interval for regulation. So far so good.

image

With Shelly LNM I can have some spikes of more than 10s without data. I will have to check why but this allowed me to test failsafe: I set a value of 10 sec (I will change the default to 15). So we see that the inverters entered failsafe mode - were caped at their minimal production, then production went up again.

I definitely need to raise the failsafe to 15-20 sec.

> Regarding current dead zone, it seems very small.

Which is good : it means your regulation is working well and is able to oscillate only a little around the setpoint. A "dead zone" would introduce oscillations. Adjustments are only each 10 sec so that's acceptable.

elmr91:
I lost communication with inverters twice this morning with beta12 with my default -7dBm.

Should be related to  https://github.com/tbnobody/OpenDTU/pull/3101
I need to change signal level with this PR applied to get 100% reliability:
-7 dBm works fine without PR
+7 dBm works fine with this PR

mathieucarbou:
> I lost communication with inverters twice this morning with beta12 with my default -7dBm.
> 
> Should be related to #3101 I need to change signal level with this PR applied to get 100% reliability: -7 dBm works fine without PR +7 dBm works fine with this PR

That might be good feedback to add as comment of #3101 !

Here everything works fine for now. I still see with Shelly LNM sometimes some measurement not received, so failsafe mode gets enables for a short time. I added a diagnostic sensor to tack the grid power age since last time it was received. It will be in next release.

image

Other wise when I plugged the EV car for charging, setpoint was set to -1000, opendtu reacted to give more power and charge started.

image

elmr91:
One test with 2000W load:
image


Target is set at -800W
Load start +2kW, 1200W taken from Grid

Solar production increased (10s) +500 W (1st round)
Solar production increased (10s) +350 W (2d round)
it was the max solar can produce....
Limit continues to increase to reach 2900 W


When load stopped, limit was adjusted in 2 rounds.

mathieucarbou:
Today: 

image

- production is ramping up, regulation ensures the production limit is targeting -200W (my setpoint).

- EV car plugged, OpenEVSE starts charging and try to take as much power as it can, regulation automatically increases the production limit and sets it to 30% more than the current solar production to give some room for increases without allowing too much space to speed up decreases

- 5 PM: zero-export turned off: no effect since EV charging is still on and makes sure to only charge on the excess power

- Some communication issues: shelly lnm paquets delayed so failsafe mode activated for a short time. I pushed a fix to set to 300 seconds (5min) the failsafe timeout. By default it was too short and was activating failsafe mode for transiant communication issues.

**Conclusion**

Everything is working pretty well!

[v26.3.30-mathieu14](https://github.com/mathieucarbou/tbnobody-OpenDTU/releases/tag/v26.3.30-mathieu14) was releases which brings support for Shelly 3EM both in 3-phase and single-phase use cases.

mathieucarbou:
New release: [v26.3.30-mathieu15](https://github.com/mathieucarbou/tbnobody-OpenDTU/releases/tag/v26.3.30-mathieu15)

Bugfix for MQTT and clear separation between the persisted config used at startup and the live config (enable and setpoint) used at runtime that can be changed through Home Asisstant

image

image

PR #3230, opened 2026-09-20
Fix: console log lines misclassified as warnings

mathieucarbou:
The virtual debug console classified each log line by searching for the level tag ('E (', 'W (', 'I (', ...) anywhere in the line. Since 'W (' is checked before 'I (', any info line whose message text contains a watt value followed by a parenthesis — e.g.

  I (122037) zeroexport: Inverter 1410A01ED5A9: 370 -> 372 W (max 1000 W)

— was matched by 'W (' and rendered in the warning color.

ESP-IDF log lines always start with the level letter ('I (time) TAG: message'), so anchor the classification on the start of the line using startsWith() instead of includes().

Also rebuild webapp_dist.


tbnobody:
Could you please remove the generated bundle from the commit  (and force push changes)? I only generate the bundles before the release otherwise it will lead to merge commits.

mathieucarbou:
> Could you please remove the generated bundle from the commit (and force push changes)? I only generate the bundles before the release otherwise it will lead to merge commits.

done ;-)

tbnobody:
Thank you. Will add it to the next release.

PR #3231, opened 2026-09-21
fix(hoymiles): guard empty command queue during deduplication

mathieucarbou:
ActivePowerControlCommand uses QueueInsertType::RemoveOldest. When such a command is enqueued while the radio command queue is empty, CommandQueue::removeDuplicatedEntries() previously evaluated:

    _queue.begin() + 1

For an empty std::deque, begin() equals end(). Advancing the past-the-end iterator by one is outside the valid iterator range and results in undefined behavior. On the ESP32 this could corrupt queue state or hang the scheduler, appearing as a deadlock: MQTT, WebSocket updates, and serial logs stopped while the REST API remained responsive.

Zero-Export made the bug easier to trigger because its failsafe, release, and keep-alive paths enqueue ActivePowerControlCommand instances when the grid source becomes unavailable. The same issue was also reachable from manual Web/API limit commands and inverter limit retries.

Guard both RemoveOldest and ReplaceExistent queue operations before using begin() + 1.

References:
- C++ iterator range requirements: https://eel.is/c++draft/iterator.requirements.general
- C++ random-access iterator requirements: https://eel.is/c++draft/iterator.concept.random.access
- std::deque::begin() behavior for an empty deque: https://en.cppreference.com/w/cpp/container/deque/begin


tbnobody:
Thank you. Will merge in next release

tbnobody:
I just saw that this is a duplicate of #3164
I would merge the "original" one.

mathieucarbou:
oh boy! I didnt see it and bumped into this issue for a while lol ;-)

